PR #9, opened 2017-12-15
chore: dependency updates and tests improvements

bartoszmajsak:
- Bumped most of the dependencies besides Arquillian. It seems that Shrinkwrap Resolver changed how the dependencies are fetched and with newest version (3.0.x, vs 3.0.0-alpha-4 we have here) there are classloading issues while deploying tests archives. I will open issue about it
- Improved few tests by using `assertj` and `rest-assured`
- Aligned naming for `version.` properties in poms

### DIFF
--- a/services/openshift-service-impl/src/test/java/io/fabric8/launcher/service/openshift/impl/OpenShiftServiceIT.java
+++ b/services/openshift-service-impl/src/test/java/io/fabric8/launcher/service/openshift/impl/OpenShiftServiceIT.java
@@ -1,18 +1,5 @@
 package io.fabric8.launcher.service.openshift.impl;
 
-import java.io.File;
-import java.io.InputStream;
-import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.util.Collections;
-import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-import javax.inject.Inject;
-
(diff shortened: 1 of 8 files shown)


services/openshift-service-impl/src/test/java/io/fabric8/launcher/service/openshift/impl/OpenShiftServiceIT.java
```diff
@@ -186,7 +187,9 @@ public void listProjects() {
 
         //then
         assertNotNull(projects);
-        assertThat(projects).extracting(OpenShiftProject::getName).allMatch(s -> s.startsWith(PREFIX_NAME_PROJECT));
+        assertThat(projects).extracting(OpenShiftProject::getName)
+                .describedAs("expecting to match " + PREFIX_NAME_PROJECT + "*")
+                .allMatch(s -> s.startsWith(PREFIX_NAME_PROJECT));
```
gastaldi: We have a known issue that this assertion fails if the OS cluster has other projects. Can you fix that?

bartoszmajsak: You mean `myproject`? I just added a description here, but sure.

gastaldi: Yes, exactly

PR #13, opened 2017-12-18
feat(tests): brings in-memory git server for LauncherIT

bartoszmajsak:
This can only be merged when we release `booster-catalog-service`.

Besides that, the code is ready to be merged from my point of view. Opening now to gather feedback and discuss the approach.


addons/launcher-addon/pom.xml
```diff
@@ -11,7 +11,7 @@
   <artifactId>launcher-addon</artifactId>
 
   <properties>
-    <version.booster-catalog-service>14</version.booster-catalog-service>
+    <version.booster-catalog-service>15-SNAPSHOT</version.booster-catalog-service>
```
gastaldi: This needs to be changed to `15`

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/BoosterCatalogFactory.java
```diff
@@ -33,15 +32,15 @@
  */
 @ApplicationScoped
 public class BoosterCatalogFactory {
-    public static final String CATALOG_GIT_REPOSITORY_PROPERTY_NAME = "LAUNCHER_BACKEND_CATALOG_GIT_REPOSITORY";
+    private static final String LAUNCHER_BOOSTER_CATALOG_REPOSITORY = "LAUNCHER_BOOSTER_CATALOG_REPOSITORY";
```
gastaldi: These system properties are set in https://github.com/fabric8-launcher/launcher-backend/blob/master/openshift/template.yaml and also in https://github.com/fabric8-launcher/launcher-openshift-templates/blob/master/openshift/launcher-template.yaml. 

Make sure to change that too.

gastaldi:
Also, booster-catalog-service is used in LaunchResource so I guess it makes sense to move the `version.booster-catalog-service` maven property out of `addons/launcher-addon`and declare this library in the `dependencyManagement` section in the parent `pom.xml`

gastaldi:
And the readme file too. Honestly I think we should leave the system env names unchanged :)

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/BoosterCatalogFactory.java
bartoszmajsak: Good spot, thank you!

bartoszmajsak:
> And the readme file too. Honestly I think we should leave the system env names unchanged :)

That's easy to fix. No worries I got it covered :)

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/BoosterCatalogFactory.java
bartoszmajsak: Template updated and PR open https://github.com/fabric8-launcher/launcher-openshift-templates/pull/35

bartoszmajsak:
> And the readme file too. Honestly I think we should leave the system env names unchanged :)

I don't see any relevant properties mentioned in the README at this point - do we want to add them?

bartoszmajsak:
Lucky PR#13 ready to be merged ;)

pom.xml
```diff
@@ -123,6 +124,21 @@
         <version>${project.version}</version>
         <classifier>forge-addon</classifier>
       </dependency>
+      <dependency>
+        <groupId>io.openshift</groupId>
```
gastaldi: The idea was not to add booster-catalog-service as a dependency in root, but in the dependencyManagement section and as a dependency in web and the launcher-addon without the version.

The other modules shouldn't depend on this library

bartoszmajsak:
Let's wait for @edewit with the review and I can take care of the merge. Thanks!

waiting

quintesse:
I think the ConfigMap key names should be aligned with the env names.

As an aside; why does one name mention "BOOSTER" and the other one doesn't? Doesn't seem very regular?

edewit:
oops sorry, didn't know people where waiting on me I'm so :+1: on this tired of cleaning up test repos from my github.

bartoszmajsak:
> I think the ConfigMap key names should be aligned with the env names.

@quintesse I tried to align them https://github.com/fabric8-launcher/launcher-openshift-templates/pull/35/files and in the template here. Did I miss something?

> As an aside; why does one name mention "BOOSTER" and the other one doesn't? Doesn't seem very regular?

This can be easily fixed. We either make it everywhere with `BOOSTER` or remove from one of them - what do you prefer?

bartoszmajsak:
@quintesse ah, you mean `launcher.backend.catalog.git.repository` etc. Sure, let me fix that.

gastaldi:
Remember that any changes to configmap keys must be done after the SD team changes in prod-preview and production to avoid deployment failures

bartoszmajsak:
Let's take this discussion offline.

bartoszmajsak:
The reason for fiddling with env variables was following - we have both Launcher Backend and Booster Catalog Service relying on url to git repository and it was all mixed, so in order to make those tests running locally you had to specify way more properties than needed, e.g.

`LAUNCHPAD_BACKEND_CATALOG_GIT_REPOSITORY` and `LAUNCHER_BACKEND_CATALOG_GIT_REF` and in fact it's not any backend catalog we are talking about here, but rather the catalog of boosters, hence the reason for change. But I think all the other changes were a bit premature and resulted in slight inconsistency. 

So we have two options here - either we roll back to old names, but unify again the booster-catalog service we use here (new release, but I can do some more boyscoutting there to make it more meaningful ;)), or we simplify the configs but then we will have to take care of  all the things around mentioned in this PR (coordinate with SD, take care of templates and so on). Preferences @gastaldi @gastaldi @quintesse 

It's a bit bikeshedding discussion we are having now, but it's my fault so I take the blame.

gastaldi:
I'd rather not bother the SD team before January unless you can convince them that this is critical. 

I'd recommend to change the env vars only and keep the configmap keys untouched. At least for now.

quintesse:
Agreed let's only change the env vars for now. I want to change those
ConfigMap keys anyway. (Because there are some nice options when you make
then exactly the same as the env vars themselves, instead of choosing
different ones like we do now)

gastaldi:
@edewit @quintesse mind reviewing this?

gastaldi:
Thanks!

PR #18, opened 2018-01-11
fix(test): adds missing AbstractGitService

bartoszmajsak:
It seems that integration tests were broken. Hence I have two questions:

- why do we have them if we don't run them as a part of the build?
- shouldn't `mvn clean install` run them by default? What's the benefit of `it` profile?

### DIFF
--- a/services/git-service-impl/src/test/java/io/fabric8/launcher/service/github/impl/GitHubServiceIT.java
+++ b/services/git-service-impl/src/test/java/io/fabric8/launcher/service/github/impl/GitHubServiceIT.java
@@ -1,28 +1,12 @@
 package io.fabric8.launcher.service.github.impl;
 
 
-import java.io.File;
-import java.net.HttpURLConnection;
-import java.net.URI;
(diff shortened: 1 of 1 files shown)


gastaldi:
You can totally blame me for that. I performed the changes and haven't run the IT tests :P

gastaldi:
Thanks !

bartoszmajsak:
> You can totally blame me for that. I performed the changes and haven't run the IT tests :P

I don't want to blame anyone - I want to fix this. Everyone is making mistakes and we have things like tests (and compilers ;)) to catch them.

alexeykazakov:
The tests must be the part of the CI job triggered by GitHub PRs + part of CD so even committed code can't be deployed if they fail.

bartoszmajsak:
In order to run then we need to talk to openshift cluster (let it be local or remote) which makes it a bit tricky (depends on where we want to run those builds though). Plus bunch of other things.

@gastaldi let's take it offline and discuss how we could tackle it.

PR #19, opened 2018-01-15
feat: introduces in-memory git rules for BoosterCatalogFactoryTest

bartoszmajsak:
Fixes https://github.com/arquillian/arquillian-organization/issues/1


addons/launcher-addon/src/test/java/io/fabric8/launcher/addon/BoosterCatalogFactoryTest.java
```diff
@@ -38,8 +51,7 @@ public void testDefaultCatalogServiceNotNullAndIsSingleton() {
 
     @Test
     public void testMasterCatalogIsNotSameAsDefault() {
-        // A null catalogURL means use default repository URL
-        BoosterCatalog masterService = factory.getCatalog(null, "master");
```
gastaldi: I think this test is supposed to test what happens when `null` is passed to the `factory.getCatalog` method, no?

bartoszmajsak: Not sure how this got lost. Maybe during rebase.

gastaldi: The idea is to prove that if a `null` is passed in that argument, the default catalog is returned (the same as `getDefaultCatalog()`.

bartoszmajsak: Then it has to talk to github as per current state of the code (based on constant values if `null` is passed). Shall we discard this PR then? Or do we want to introduce e.g. `DefaultCatalogProvider` which we could substitute for this test instead of using `DEFAULT_GIT_REPOSITORY_URL` ?

gastaldi: I think we already test with an in-memory Git server in the launcher-booster-catalog-service repository, right? The main purpose of this test here is to verify if the expected objects are being returned in the public methods. What would this `DefaultCatalogProvider` do? Perhaps If we did lazy indexing, I don't think we'd need a whole lot of infrastructure just to check if the returned objects match.

bartoszmajsak: I think with https://github.com/fabric8-launcher/launcher-backend/commit/adabb8c9311bc6e0ad542ed66f24cca71b8fa95f we are good with the intended behaviour of the test. Good to merge @gastaldi?

gastaldi:
Thanks!

PR #20, opened 2018-01-15
fix: makes use of in-memory git server

bartoszmajsak:
As we already have in-memory git server in place we can local git repositories while running tests instead of talking with github. This PR brings the use of it for `HealthResourceIT` as we are loading boosters while deploying the whole app.

### DIFF
--- a/web/src/test/java/io/fabric8/launcher/web/api/HealthResourceIT.java
+++ b/web/src/test/java/io/fabric8/launcher/web/api/HealthResourceIT.java
@@ -15,8 +15,16 @@
  */
 package io.fabric8.launcher.web.api;
 
-import java.io.StringReader;
-import java.net.URI;
+import org.arquillian.smart.testing.rules.git.server.GitServer;
+import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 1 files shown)


web/src/test/java/io/fabric8/launcher/web/api/HealthResourceIT.java
```diff
@@ -43,6 +45,21 @@
 @RunWith(Arquillian.class)
 public class HealthResourceIT {
 
+    @ClassRule
+    public static GitServer gitServer = GitServer.bundlesFromDirectory("repos/boosters")
```
gastaldi: Wouldn't be better if this code stood in a common superclass? Or perhaps the @Deployment is adding too much instead of just adding HealthResource?

bartoszmajsak: I am not sure if we gain a lot from such a re-use. Afterall it's few lines of code vs making a constraint by introducing some "common" superclass.

But overall yes - it's deploying the whole app which might not be what we want.

gastaldi:
Merged, thanks!

PR #21, opened 2018-01-17
can also be an env property

edewit:
`Boolean.getBoolean` only uses the `System.getProperty`, but once deployed `"LAUNCHER_SKIP_OOF_CATALOG_INDEX"` is a environment variable


gastaldi:
Can we use `EnvironmentSupport` instead? Also I don't see this flag set in the template, so I assume this env var is used only when testing locally?

PR #25, opened 2018-01-18
Introduced Gitlab integration

gastaldi:
This is part of https://github.com/fabric8-launcher/launcher-planning/issues/93


gastaldi:
According to https://docs.gitlab.com/ee/api/README.html#authentication, the GitLab API offers two options of tokens. This implementation uses the  [Personal Access Token](https://docs.gitlab.com/ee/api/README.html#personal-access-tokens) strategy

gastaldi:
Any feedback? I'd like to merge this

base/src/test/java/io/fabric8/launcher/base/identity/IdentityFactoryTest.java
```diff
@@ -15,6 +15,7 @@
     public void testTokenIdentity() {
         TokenIdentity identity = IdentityFactory.createFromToken("FOO");
         Assert.assertThat(identity.getToken(), equalTo("FOO"));
+        Assert.assertThat(identity.getType().isPresent(), equalTo(false));
```
bartoszmajsak: Can we use `assertj` assertions instead? I believe some of the tests are already based on that. Makes it nicer to read and easier to write.

pom.xml
```diff
@@ -277,6 +277,13 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.immutables</groupId>
+      <artifactId>value</artifactId>
+      <version>2.5.5</version>
```
bartoszmajsak: Can we make the version as a property?

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitUser.java
```diff
@@ -1,14 +1,18 @@
 package io.fabric8.launcher.service.git.api;
 
+import org.immutables.value.Value;
+
 /**
  * Value Object representing a Git user
  *
  * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
  */
+@Value.Immutable
```
bartoszmajsak: How much do we gain from this? If it's only about properly made value objects is it really worth an effort to have yet another lombok-inspired annotation processor? Does it have IDE support or do I have to re-run the mvn build to see the changes?

services/git-service-impl/src/main/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceFactoryImpl.java
```diff
@@ -0,0 +1,36 @@
...
+            throw new IllegalArgumentException("GitLabService supports only TokenIdentity. Not supported:" + identity);
+        }
+        return new GitLabServiceImpl((TokenIdentity) identity);
+    }
+
+    @Override
+    public Optional<Identity> getDefaultIdentity() {
+        // Try using the provided Github token
+        String token = EnvironmentSupport.INSTANCE.getEnvVarOrSysProp(GITLAB_PRIVATE_TOKEN);
+        return token == null ? Optional.empty() : Optional.of(IdentityFactory.createFromToken("Private-Token", token));
```
bartoszmajsak: sth like `return Optional.ofNullable(token).map(token -> IdentityFactory.createFromToken("Private-Token", token))` ? but that's more a matter of taste I guess.

services/git-service-impl/src/main/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceImpl.java
```diff
@@ -0,0 +1,207 @@
...
+    @Override
+    public GitRepository createRepository(String repositoryName, String description) throws IllegalArgumentException {
+        StringBuilder content = new StringBuilder();
+        content.append("name=").append(repositoryName);
+        if (description != null && !description.isEmpty()) {
+            content.append("&description=").append(description);
+        }
+        Request request = request()
+                .post(RequestBody.create(APPLICATION_FORM_URLENCODED, content.toString()))
+                .url("https://gitlab.com/api/v4/projects")
```
bartoszmajsak: Can we make URL configurable? This way we would support on-premise gitlab installation.

services/git-service-impl/src/main/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceImpl.java
```diff
@@ -0,0 +1,207 @@
...
+                .build();
+        String urlAsString = url.toString();
+        return execute(request, (JsonNode tree) -> {
+            for (JsonNode node : tree) {
+                if (urlAsString.equalsIgnoreCase(node.get("url").asText())) {
+                    return readHook(node);
+                }
+            }
+            return null;
+        }).orElse(null);
```
bartoszmajsak: Why not `Optional` instead?

services/git-service-impl/src/test/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceIT.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    public void createHook() throws Exception {
+        GitRepository repo = createRepository("my-awesome-repository", "Created from integration tests");
+        GitHook hook = gitLabService.createHook(repo, new URL("http://my-hook.com"),
+                                                GitLabWebHookEvent.PUSH.name(), GitLabWebHookEvent.MERGE_REQUESTS.name());
+        assertThat(hook).isNotNull();
+        assertThat(hook.getName()).isNotEmpty();
+        assertThat(hook.getUrl()).isEqualTo("http://my-hook.com");
+        assertThat(hook.getEvents()).containsExactly("push", "merge_requests");
```
bartoszmajsak: If you make it [soft assertion](http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#soft-assertions) you will collect all errors at once instead of re-running the test and discovering one by one. There's even JUnit rule for that.

gastaldi:
The following env vars need to be defined to run the integration tests:
```
export GITLAB_PRIVATE_TOKEN=<GITLAB_TOKEN>
export GITLAB_USERNAME=<GITLAB_USER>
```

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitUser.java
gastaldi: It's only about creating the DTOs without requiring us to maintain it. Yes, it does have IDE support, see http://immutables.github.io/apt.html

gastaldi:
I think it would be awesome if we could test against a Docker container instead. Can you help us set that up? I also agree that IT tests should always be part of the build.

Issue #28, opened 2018-01-19
Adjustments to the new Booster Catalog API


quintesse:
See fabric8-launcher/launcher-planning#65

PR #29, opened 2018-01-19
Merge booster yaml into catalog

quintesse:
See https://github.com/fabric8-launcher/launcher-planning/issues/65 and fabric8-launcher/launcher-backend#28 for more information.

You'll need to set the following values for local testing:

```
export LAUNCHER_BOOSTER_CATALOG_REPOSITORY=https://github.com/quintesse/launcher-booster-catalog.git
export LAUNCHER_BOOSTER_CATALOG_REF=staging
```


addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/ui/booster/ProjectInfoStep.java
```diff
@@ -214,9 +219,7 @@ public Result execute(UIExecutionContext context) throws Exception {
             if (runtimeVersion.getValue() != null) {
                 profileId = runtimeVersion.getValue().getId();
             }
-            if (booster.getBuildProfile() != null) {
-                profileId = booster.getBuildProfile();
-            }
+            booster.getMetadata("buildProfile", profileId);
```
gastaldi: Shouldn't the return of this method be assigned to `profileId` ?

quintesse: Good catch, thx!

gastaldi:
What changes are needed to be performed by the SD team? Can you open a housekeeping issue asking them to update the configmap values?

quintesse:
I'm too sleepy right now :)

Probably nothing more than changing 2 env vars:

```
LAUNCHER_BOOSTER_CATALOG_REF=staging
LAUNCHER_BACKEND_ENVIRONMENT=staging
```

(We'd first merge my PR for the catalog into "master" and then push a copy to "staging". That way we'd still have "next" around if something goes wrong. If everything works okay we can remove "next" at a later date)

gastaldi:
@quintesse I've been thinking a lot about this and I prefer the RHOAR classes to live in booster-catalog-service under a `rhoar` subpackage.

The reason is that I want that library to be the only dependency needed in case anyone wants to do anything different with the catalog (like the doc booster sync app  that @edewit created in the past)

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/ui/booster/DeploymentType.java
```diff
@@ -0,0 +1,32 @@
...
+ * Licensed under the Eclipse Public License version 1.0, available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package io.fabric8.launcher.addon.ui.booster;
+
+/**
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
+ */
+public enum DeploymentType {
```
gastaldi: This is ok to be here. It will be gone after we refactor the current backend

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/catalog/RhoarBooster.java
```diff
@@ -0,0 +1,75 @@
+package io.fabric8.launcher.addon.catalog;
+
+import java.util.Map;
+
+import io.openshift.booster.catalog.Booster;
+import io.openshift.booster.catalog.BoosterFetcher;
+
+public class RhoarBooster extends Booster {
```
gastaldi: I´d rather have everything in the `catalog` package as a `rhoar` sub package in launcher-booster-catalog-service

gastaldi:
Don't forget to change the OpenShift template too.

PR #32, opened 2018-01-24
fix(tests): integration tests fixes

bartoszmajsak:
Again, integration tests were not running successfully for me after latest changes on `master`. This revealed two issues:

- GitHub accounts not always have email available publicly - thus testing, or even assuming it's always there should not take place
- GitLab API for getting created project wasn't working when using `groups`. `users` resource should be used instead. This problem in fact leads to two follow-up questions
  - is our abstraction with organizations good enough? There is no notions of groups on GitHub nor organizations in GitLab and I'm not sure if we can treat them as the same concepts. Maybe the abstraction we are building is not good / needed here?
  - instead of plain http calls against GitLab maybe we can consider switching to one of the Java clients [[1]](https://github.com/timols/java-gitlab-api) [[2]](https://github.com/gmessner/gitlab4j-api)

Ah yeah, sorry for the imports again. Somehow the settings didn't persist. That's the last time.

### DIFF
--- a/services/git-service-impl/src/main/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceImpl.java
+++ b/services/git-service-impl/src/main/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceImpl.java
@@ -1,18 +1,5 @@
 package io.fabric8.launcher.service.gitlab.impl;
 
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.net.URL;
-import java.net.URLEncoder;
-import java.util.ArrayList;
-import java.util.Collections;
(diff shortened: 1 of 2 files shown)


gastaldi:
>is our abstraction with organizations good enough? There is no notions of groups on GitHub nor organizations in GitLab and I'm not sure if we can treat them as the same concepts. Maybe the abstraction we are building is not good / needed here?

I think so, afaik GitHub Organizations = Gitlab Groups. Do you think we should change this abstraction?

>instead of plain http calls against GitLab maybe we can consider switching to one of the Java clients

I considered that initially, but I wasn't sure how stable was those clients so I simply ignored it. I figured it would be easier if I just made the plain http calls with the libraries that were already there instead of adding more client libraries for each implementation.

PR #53, opened 2018-01-31
Read Jenkins URL webhook from env

gastaldi:


addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
```diff
@@ -359,16 +359,18 @@ public Result execute(UIExecutionContext context) {
...
-            try {
-                jenkinsUrl = KubernetesHelper.getServiceURL(kubernetesClientHelper.getKubernetesClient(), ServiceNames.JENKINS, jenkinsNamespace, "https", true);
-                discoveryNamespace = jenkinsNamespace;
-            } catch (Exception e) {
-                if (!discoveryNamespace.equals(jenkinsNamespace)) {
-                    try {
-                        jenkinsUrl = KubernetesHelper.getServiceURL(kubernetesClientHelper.getKubernetesClient(), ServiceNames.JENKINS, discoveryNamespace, "https", true);
-                    } catch (Exception e2) {
-                        throw new BadTenantException("Failed to find Jenkins URL in namespaces " + discoveryNamespace + " and " + jenkinsNamespace + ": " + e, e);
+            String jenkinsUrl = System.getenv("JENKINS_URL");
```
hferentschik: So that I understand this correctly, this _jenkinsUrl_ will later then become the base URL for the webhook? A new user gets provisioned, logs in, steps through the codebase creation wizard, selects "enable webhooks" and then will end up with a webhook in the repo of the form: https://\<jenkinsUrl>/github-webhook/?

hferentschik: Would JENKINS_PROXY_URL not be a better name? @vpavlin WDYT?

hferentschik: Can we get this into pro-preview? Or is it already there?

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
```diff
@@ -359,16 +359,18 @@ public Result execute(UIExecutionContext context) {
...
+            if (jenkinsUrl == null) {
```
hferentschik: hmm, that seems odd to me. So we are saying that we pass the URL via the environment, but then here you try to discover it if it is not set. Is this really a good idea? To me, JENKINS_PROXY_URL becomes a required configuration parameter and if it is not set, things should fail. Is there nothing somewhere which would verify expected and required environment settings as the container comes up. I'd argue that the container should not even start without.

vpavlin: Launcher is a general application/service not tightly coupled with OSIO, thus having proxy running is not a requirement for Launcher to be useful. Only in case it is set, it should use it - i.e. I am fine with this

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
vpavlin: > will end up with a webhook in the repo of the form:

Correct

I'd go with `JENKINS_URL` - if you think about another potential case - Launcher could be deployed in environment where there is one common Jenkins for all projects - ours is a special case where the Jenkins is not actually Jenkins but a proxy going to a random Jenkins.

vpavlin: About to push new config map to prod-preview

hferentschik: > I'd go with JENKINS_URL

Ok

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
hferentschik: Hmm, not sure about this. I'd argue JENKINS_URL should just point to whatever Jenkins instance is used. Having this fallback is a potential gateway to an error or at least unexpected behaviour. 

At the very least it should be clearly logged how the Jenkins URL was discovered.

hferentschik: Out of interest, why has everything to happen in a single method in this class. _ execute_ is a hell of a method to take in. What's about a `public String determineJenkinsUrl() throws Exception`?

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
```diff
@@ -359,16 +359,18 @@ public Result execute(UIExecutionContext context) {
...
+                try {
+                    jenkinsUrl = KubernetesHelper.getServiceURL(kubernetesClientHelper.getKubernetesClient(), ServiceNames.JENKINS, jenkinsNamespace, "https", true);
+                    discoveryNamespace = jenkinsNamespace;
+                } catch (Exception e) {
+                    if (!discoveryNamespace.equals(jenkinsNamespace)) {
+                        try {
+                            jenkinsUrl = KubernetesHelper.getServiceURL(kubernetesClientHelper.getKubernetesClient(), ServiceNames.JENKINS, discoveryNamespace, "https", true);
+                        } catch (Exception e2) {
+                            throw new BadTenantException("Failed to find Jenkins URL in namespaces " + discoveryNamespace + " and " + jenkinsNamespace + ": " + e, e);
```
hferentschik: BadTenantException!? What's that for an exception Type?

hferentschik: Should _jenkinsURL_ be validated?

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
gastaldi: This code is part of the old fork and will be removed in our refactoring. That said, I prefer to not introduce any major change in this class. I'll definitely use your feedback in the refacforing though

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
gastaldi: I know right? This code is a piece of :hankey: :)
I'll hopefully get it rewritten in a maintainable format soon. I'd like to avoid changing this part for now.

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
hferentschik: Sure. Up to you. Personally, I am a believer in the scout rule "always leave the campground cleaner as you found it". I find an extracted method cleaner and probably even less invasive. 

History proves that one often have to deal longer with these types of legacy code bases as one thinks.

addons/osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java
hferentschik: > I know right?

Yeah, nothing to write home about ;-)

PR #67, opened 2018-02-08
Fix environment script

ia3andy:
When using 'source ...' the parameter $0 is not working.
We use $BASH_SOURCE to get the file location instead.


quintesse:
Although I must admit I always doubt what to use. It's a change in documentation so it's logical to use `docs(README)`, but it's also a fix of existing documentation (and in this case of a code example inside the docs) so `fix(README)` seems correct as well. 😄

quintesse:
Perfect! Thanks @ia3andy

PR #71, opened 2018-02-09
feat: Adds service virtualization for GitLab Integration Tests

dipak-pawar:
This PR brings service virtualization based on Hoverfly to set us free from 3rd party calls during tests (in this case GitLab).


gastaldi:
That's awesome @dipak-pawar! Thank you! 

One question: Do you think we could make the GithubServiceIT lighter as this test case is (not using Arquillian)? 

Mind providing another PR with that?

services/git-service-impl/src/test/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceIT.java
```diff
@@ -1,28 +1,48 @@
...
+import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.contrib.java.lang.system.ProvideSystemProperty;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.TestName;
 
-import static io.fabric8.launcher.service.gitlab.api.GitLabEnvVarSysPropNames.LAUNCHER_MISSIONCONTROL_GITLAB_USERNAME;
+import static io.fabric8.launcher.service.github.test.HoverflyRuleConfigurer.createHoverflyProxy;
```
gastaldi: Time to move this class out of the github package

services/git-service-impl/src/test/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceIT.java
```diff
@@ -1,28 +1,48 @@
...
 
 /**
  * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
  */
 public class GitLabServiceIT {
 
+    private static final String MY_GITLAB_REPO_DESCRIPTION = "Test project created by Integration Tests.";
+
+    @ClassRule
+    public static RuleChain ruleChain = RuleChain
```
gastaldi: Awesome! How can we reuse this set of rules in other tests (like GithubServiceIT)?

dipak-pawar: RuleChain is provided by Junit - https://junit.org/junit4/javadoc/4.12/org/junit/rules/RuleChain.html.

We are using it to make sure we have system properties available to use in another rule. You can chain any Rule which implements `TestRule` interface

gastaldi: Right, I think my question was more to: How can we reuse these lines of code to avoid copy pasting in every test that uses Hoverfly. A base class?

dipak-pawar:
> Do you think we could make the GithubServiceIT lighter as this test case is (not using Arquillian)?

In GitHubServiceIT we are using arquillian, which is executing all tests inside wildfly swarm container as same like production environment. So I would keep it the same.

gastaldi:
Yes, but to be honest I don't see much difference from the Gitlab implementation, but that's ok.

services/git-service-impl/src/test/java/io/fabric8/launcher/service/gitlab/impl/GitLabServiceIT.java
dipak-pawar: yes we can do it by writing `@ClassRule` in baseclass and extending it in testclass.

gastaldi:
Retest this please

gastaldi:
Restest this please

gastaldi:
Retest this please

gastaldi:
add to whitelist

gastaldi:
Retest this please

PR #85, opened 2018-02-19
fix(README): Make documentation clearer about unsetting credentials

ia3andy:
Make documentation clearer about unsetting credentials to use KeyCloak server

It was not clear that Openshift and Git credentials needed to be unset in order to use a KeyCloak server.
Launcher environment is now displayed when sourcing the script to ensure everything is set as expected.


gastaldi:
Add to whitelist

PR #88, opened 2018-02-20
fix(README): Fix environment script directory issue

ia3andy:
dirname "$BASH_SOURCE" is relative so it's always "./"


quintesse:
Thanks Andy

PR #89, opened 2018-02-21
introduced provider renamed to OpenShiftService

edewit:


addons/new-osio-addon/src/main/java/io/fabric8/launcher/osio/providers/OpenShiftServiceProvider.java
```diff
@@ -0,0 +1,51 @@
...
+import io.fabric8.launcher.osio.EnvironmentVariables;
+import io.fabric8.launcher.osio.openshift.OpenShiftService;
+import io.fabric8.launcher.osio.tenant.Tenant;
+
+import static io.fabric8.launcher.core.spi.Application.ApplicationType.OSIO;
+
+@RequestScoped
+public class OpenShiftServiceProvider {
+
+    @Produces
```
quintesse: @edewit: I saw this `@Produces` annotation without any mime-types in @gastaldi 's code as well, but why is it here? According to what I read in the docs this annotation a) needs an argument and b) it's to decide what to result to return to the browser/rest client that accesses it. But this is an internal service that will never return any information to any client, so why do we mark it with this annotation?

edewit: I'm using `javax.enterprise.inject.Produces` here and the documentation that you refer to is `javax.ws.rs.Produces`. Used here it marks the method as a factory method that needs to be invoked by CDI to create the `OpenShiftService` for further injection

quintesse: Aaah! You're right of course, thx, I must have looked at one of the WS ones and assumed they were all the same. Wow, that's really confusing.

addons/new-osio-addon/src/main/java/io/fabric8/launcher/osio/providers/OpenShiftServiceProvider.java
```diff
@@ -0,0 +1,51 @@
...
+    @Application(OSIO)
+    public OpenShiftService createOpenShiftService(HttpServletRequest request, IdentityProvider identityProvider, Tenant tenant) {
+        String openShiftApiURL = EnvironmentVariables.getOpenShiftApiURL();
+        String authorization = request.getHeader(HttpHeaders.AUTHORIZATION);
+
+        ConfigBuilder configBuilder = new ConfigBuilder();
```
gastaldi: Why noy use the OpenshiftServiceFactory here?

addons/new-osio-addon/src/test/java/io/fabric8/launcher/osio/steps/OpenShiftStepsTest.java
```diff
@@ -82,10 +76,17 @@ public void shouldCreateBuildConfig() throws IOException, URISyntaxException {
...
+        Tenant tenant = ImmutableTenant.builder().identity(identity)
                 .username("edewit").email("me@nerdin.ch").namespaces(elements).build();
 
-        steps.tenant = openshiftClient.tenant;
+        String openShiftApiURL = EnvironmentVariables.getOpenShiftApiURL();
+        Config config = new ConfigBuilder().withMasterUrl(openShiftApiURL).withOauthToken("123")
+                .withTrustCerts(true).build();
+
+        OpenShiftService openShiftService = new OpenShiftService(new DefaultKubernetesClient(config), tenant);
+        steps.openShiftService = openShiftService;
```
gastaldi: This looks like a code smell. Is this because Arquillian is too slow to run this test?

gastaldi:
Replaced by https://github.com/fabric8-launcher/launcher-backend/pull/126. Closing

PR #103, opened 2018-02-26
added a note in readme

animuk:
added a note in readme for local testing


README.md
```diff
@@ -53,6 +53,7 @@ $ ./docker.sh
 ```
 
 This will build and run a Docker image that will work in the same way as if you were running it locally. So the frontend can connect to it in the same way.
+* Please Note : for local testing, we can push the generated docker image into docker registry and use the docker registry reference in the osio dev-cluster to test the changes before merging into the master branch..
```
gastaldi: What docker registry are you referring about?

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
```diff
@@ -136,4 +136,4 @@ public Response reindex(@QueryParam("token") String token) {
         boosterCatalogFactory.reset();
         return Response.ok().build();
     }
-}
\ No newline at end of file
+}
```
gastaldi: This doesn't look like it belongs to this PR, can you remove from this PR, please?

README.md
animuk: user docker registry for local testing

PR #108, opened 2018-02-26
feat(git): Introduced /services/git/user

gastaldi:
This fixes https://github.com/fabric8-launcher/launcher-planning/issues/112 by introducing a new endpoint 
containing the necessary logged user information (login, email and avatar_url).


gastaldi:
What code cleanup are you referring to specifically?

quintesse:
Well look at the diff of your PR, 3/4 of the changes are import re-ordering, reformatting of lines and adding or removing spaces. That makes the important changes harder to read. If you're making reviews required you should at least make it easy for us :)

quintesse:
So I'd suggest to always put those in a separate commit called `chore: clean up` or something

gastaldi:
Well, the reordering of imports seems to change between our IDEs. Even if I don't change anything in the code, the import order is different. I wonder if the formatter file needs to be fixed or if I am missing something in my IDE configuration.

As for having reordering of imports in different commits that would be nice, but really unpractical. That would involve stashing partial changes and would have near-zero benefit IMHO, given that the review is performed based on all commits of a PR.

quintesse:
> Even if I don't change anything in the code, the import order is different.

I think you should start by disabling automatic code fixing in your IDE. It shouldn't be making changes you haven't asked for.

> that would be nice, but really unpractical.

I'm sorry, but I think this is unacceptable. Splitting out formatting/style changes from "real" changes is the most normal thing to do. There is a reason http://karma-runner.github.io/2.0/dev/git-commit-msg.html has a `style` type.

See: https://wiki.openstack.org/wiki/GitCommitMessages#Things_to_avoid_when_creating_commits and : https://gist.github.com/stephenparish/9941e89d80e2bc58a153#recognizing-unimportant-commits

> That would involve stashing partial changes

`git add -p` is your friend. I use it all the time. See https://johnkary.net/blog/git-add-p-the-most-powerful-git-feature-youre-not-using-yet/ and https://css-tricks.com/git-add-patch-mode/

> and would have near-zero benefit IMHO

The benefit is that the reviewer can see at a glance which things are important and which are not. You're making the reviewer do the work that _you_ should be doing.

Now, I'm not saying that a commit should be split just because a single space was added somewhere, but the moment the commit is more layout changes than anything else as in this example we should really be doing everyone a favor and split things out. Commits should be as understandable as possible.

PR #113, opened 2018-02-27
doc(README): Improve keycloack info in readme

ia3andy:
This is just a preview for review, it is still missing stuff to be merged.


ia3andy:
https://github.com/ia3andy/launcher-backend/blob/improve-keycloack-info-in-readme/README.md

README.md
```diff
@@ -57,268 +51,76 @@ This will build and run a Docker image that will work in the same way as if you
...
 ---------------------------
 
 Run the following command, replace TOKEN with the value defined in the environment variable `LAUNCHER_BACKEND_CATALOG_REINDEX_TOKEN`. Doesn't need to be specified if the environment variable doesn't exist in the running environment:
 
         $ curl -v -H "Content-Type: application/json" -d '{}' -X POST  https://localhost:8180/api/launchpad/catalog/reindex\?token\=TOKEN
+        
+        
+Setup git providers default credentials (No KeyCloak mode)
+----------------------------------------------------------
+
```
quintesse: I think we need to make it more visually obvious here that there are 3 different options. Right now it just looks like a list of 6 items and that you have to perform all of them.

quintesse: Also add a text perhaps that says that you can use any combination of them (as long as you have at least one configured).

README.md
```diff
@@ -8,20 +8,17 @@ an Angular Front application responsible to collect from an end user the informa
...
+ 
+Setting up the environment
+--------------------------
+
+
+* You have to setup environment variables before you start the back-end. 
+
+> You can source, or use as an example, the environment script template located there: [./launcher-env-template.sh](./launcher-env-template.sh)
+
+if you are not using KeyCloak, you can find information on how to [setup your git providers default credentials](README.md#setup-git-providers-default-credentials-no-keycloak-mode).
```
quintesse: Uppercase `I`

README.md
```diff
@@ -31,12 +28,9 @@ Build and Run the Application
 $ mvn clean install
 ```
 
-* Execute the application in the easiest, most generic manner with:
-```bash
-$ ./run.sh
-```
+* Source the [the environment script](README.md#setting-up-the-environment)
```
quintesse: I would just move this to the "Setting up the environment" section

launcher-env-template.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/sh
+
+#################################################
+# Here you can find basic setup for this script #
+#################################################
+
+# Choose (uncomment) one of the 3 KeyCloak options below
```
quintesse: I was thinking that we could perhaps put an `if` around this to test if this value is already set in the environment. If so we just use that one. That makes it possible to switch without having to edit this file, if that's what you prefer. Wdyt?

launcher-env-template.sh
```diff
@@ -0,0 +1,104 @@
...
+    # Official KeyCloak
+    export LAUNCHER_KEYCLOAK_URL=https://sso.openshift.io/auth
+    export LAUNCHER_KEYCLOAK_REALM=rh-developers-launch
+    export LAUNCHER_MISSIONCONTROL_OPENSHIFT_CLUSTERS_FILE=$SCRIPT_DIR/clusters.yaml
+    ;;
+"LOCAL")
+    # Local KeyCloak
+    export LAUNCHER_KEYCLOAK_URL=http://localhost:8280/auth
+    export LAUNCHER_KEYCLOAK_REALM=launch
+    export LAUNCHER_MISSIONCONTROL_OPENSHIFT_CLUSTERS_FILE=$SCRIPT_DIR/clusters.yaml
```
quintesse: I think we need to explain this. Which probably means we need a KeyCloak section in the README.

README.md
ia3andy: For now, I don't think we can use any of them, only github is plugged. It's more for tests (for now)

ia3andy:
@quintesse @gastaldi @edewit I made all the changes, it should be ok now..

gastaldi:
@centos-ci retest this please

PR #121, opened 2018-02-28
chore(env): Fixed OSIO env vars

gastaldi:


launcher-env-template.sh
```diff
@@ -99,11 +99,19 @@ export LAUNCHER_BACKEND_URL="http://127.0.0.1:8080/api" #TODO needs to be defaul
 # Testing tracker token
 export LAUNCHER_TRACKER_SEGMENT_TOKEN=dMV5AjaweCpO3KZop7TuZ0961UO74AF0
 
-# For OSIO addon in the backend
-export WIT_URL=https://api.openshift.io
-export AUTH_URL=https://auth.openshift.io
-export KEYCLOAK_SAAS_URL=https://sso.openshift.io/
-export OPENSHIFT_API_URL=https://api.starter-us-east-2.openshift.com
+# For OSIO addon in the backend - Staging
+export WIT_URL=https://api.prod-preview.openshift.io
```
ia3andy: Could we use an external variable with a switch for this as well? (like KEYCLOAK)
OSIO=PROD or STAGING

gastaldi: Good idea. Let me do that

Issue #112, opened 2018-02-27
Fix IT and then add them to CI build process


ia3andy:
There are errors in modules new-osio-addon and launcher-web.

MissionControlIT is not deleting git repository named "foo", then it fails when we re-run the tests.
(https://github.com/fabric8-launcher/launcher-backend/issues/102)

A test fail in module core-impl, the solution is:
`export OPENSHIFT_API_URL=f8osoproxy-test-dsaas-production.09b5.dsaas.openshiftapps.com`
-> try to find a more generic solution.

gastaldi:
Hoverfly needs to be updated to also consider 
```
export OPENSHIFT_API_URL=https://f8osoproxy-test-dsaas-preview.b6ff.rh-idev.openshiftapps.com
```

gastaldi:
The CI build PR to support running IT is https://github.com/fabric8-launcher/launcher-backend/pull/75

ia3andy:
I will try to fix the OPENSHIFT_API_URL bug and the "foo" github repository not deleted.

PR #123, opened 2018-02-28
fix(IT): MissionControlIT test was never deleting github repository

### LINKED ISSUE #102: MissionControlIT is not deleting git repository named "foo"
Then it fails when we re-run the tests.

ia3andy:
MissionControlIT test was never deleting github repository 'foo'

Fixes #102, #112


gastaldi:
Add to whitelist

PR #131, opened 2018-03-01
feat(TEST): Create launcher-common-test module

ia3andy:
Move test common classes to a new module for test: launcher-common-test


addons/new-osio-addon/pom.xml
```diff
@@ -25,6 +25,11 @@
     </dependency>
 
     <!-- Test dependencies -->
+    <dependency>
+      <groupId>io.fabric8.launcher</groupId>
+      <artifactId>launcher-common-test</artifactId>
```
gastaldi: launcher-common-test? But that only contains Hoverfly classes. Perhaps call it `launcher-hoverfly`?

PR #143, opened 2018-03-02
feat(hoverfly): improve hoverfly setup

### LINKED ISSUE #137: Remove LAUNCHER_TESTS_TRUSTSTORE_PATH from script
Now that the Hoverfly keystore is part of the launcher-base-test JAR, check if it's possible to point to the keystore file inside that JAR (or copy to a temp dir) instead of having to explicitly specify it in the tests

ia3andy:
Fixes #137


base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherHoverflyEnvironment.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private File createTemporaryFolderIn(File parentFolder) throws IOException {
+        File createdFolder = File.createTempFile("junit-hoverfly", "", parentFolder);
+        createdFolder.delete();
+        createdFolder.mkdir();
+        return createdFolder;
+    }
+
+    private void initTrustStore() throws IOException {
+        trustStoreTempFolder = createTemporaryFolderIn(null);
```
gastaldi: I think it's just a file, why do you need a temp folder?

ia3andy: to use copy :) which is simpler

ia3andy: But if you insist :p I would just create a file (after seeing your perf concerns :).

gastaldi: You can copy directly to the temporary folder afaik, `Paths.get(System.getProperty("java.tmp.dir"))` or something like that will give you the tmp folder path

gastaldi: but Files.createTempFile is easier, I believe :smile:

base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherHoverflyEnvironment.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public LauncherHoverflyEnvironment and(final String name, final String value) {
+        super.and(name, value);
+        return this;
+    }
+
+    private void initTrustStore() throws IOException {
+        trustStoreTempFile = File.createTempFile("hoverfly", ".jks");
+        Files.copy(LauncherHoverflyEnvironment.class.getResourceAsStream("/hoverfly/hoverfly.jks"), trustStoreTempFile.toPath(), REPLACE_EXISTING);
```
gastaldi: Afaik you need to close this `InputStream`. Use a try-with-resources block

base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherHoverflyEnvironment.java
ia3andy: My last update is doing that :) I ended up with the same conclusion..

base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherHoverflyEnvironment.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public LauncherHoverflyEnvironment and(final String name, final String value) {
+        super.and(name, value);
+        return this;
+    }
+
+    private void initTrustStore() throws IOException {
+        trustStoreTempFile = File.createTempFile("hoverfly", ".jks");
```
gastaldi: You can use `Files.createTempFile` here to avoid the `java.io.File` API

base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherHoverflyEnvironment.java
ia3andy: Ouch yeah...

PR #158, opened 2018-03-07
fix(GIT): use abstraction to test each GitService implementation

### LINKED ISSUE #100: Abstract integration tests on git services
Tests are nearly the same for all git services, it would be nice to abstract common tests.

ia3andy:
- GitService.createHook can now be called without events, it will use default ones.
- Includes multiple bug fixes on service GitService implementation to behave well according to the contract.

Fixes #100


addons/new-osio-addon/src/main/java/io/fabric8/launcher/osio/steps/GitSteps.java
```diff
@@ -80,7 +80,7 @@ public void createWebHooks(OsioProjectile projectile, GitRepository gitRepositor
         String jenkinsWebhookURL = EnvironmentVariables.ExternalServices.getJenkinsWebhookURL();
         try {
             // TODO: Check if the webhook requires a secret
-            gitService.createHook(gitRepository, null, new URL(jenkinsWebhookURL));
+            gitService.createHook(gitRepository, null, new URL(jenkinsWebhookURL), gitService.getSuggestedNewHookEvents());
```
gastaldi: Is this change really needed? The `gitService.getSuggestedNewHookEvents());` is the default value if nothing is provided, right?

ia3andy: Indeed I can remove it now it's part of the doc:
```
     * @param events     - the events that trigger the webhook; if none specified, {@see getSuggestedNewHookEvents} is used by default.
```

gastaldi:
[test]

PR #165, opened 2018-03-08
fix(GIT): Use CI credentials for hoverfly git test records

ia3andy:
Fix deleteHooks by manually setting states because it was not working correctly.


services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceIT.java
```diff
@@ -476,16 +478,29 @@ private GitRepository createRepository(GitOrganization organization, String repo
...
         return repository;
     }
 
-    private String generateRepositoryName(final int number) {
+    protected String generateRepositoryName(final int number) {
         return this.getClass().getSimpleName().toLowerCase() + "-" + testName.getMethodName().toLowerCase() + "-" + number;
     }
 
+    @Test
+    @Ignore
```
gastaldi: Why `@Ignore`d?

services/git-service-impl/src/test/java/io/fabric8/launcher/service/github/GitHubServiceIT.java
```diff
@@ -57,4 +65,22 @@ protected String getRawFileUrl(String fullRepoName, String fileName) {
         return "https://raw.githubusercontent.com/" + fullRepoName + "/master/" + fileName;
     }
 
+    @Test
+    @Ignore
```
gastaldi: Why @Ignored ? Might as well remove it?

services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceIT.java
ia3andy: Because I use it when there is an error and repositories are not deleted (instead of cleaning manually). Maybe I should add a comment though.

services/git-service-impl/src/test/java/io/fabric8/launcher/service/github/GitHubServiceIT.java
ia3andy: I could remove this one though. I used it to create a request and use it for deleteHook state

PR #170, opened 2018-03-09
feat(build): checkstyle validation when installing

### LINKED ISSUE #133: Add a checkstyle validation on PR
I don't know if it should be part of CI or another tool..

@gastaldi  said:
we can reuse what's used in the wildfly-swarm repositories

https://github.com/wildfly-swarm/checkstyle

The checkstyle config that is used by all projects in the WildFly Swarm organisation
or create our own which may be better I think

ia3andy:
Fixes #133

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -426,6 +426,24 @@
           <serverId>ossrh</serverId>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>validate</id>
+            <phase>validate</phase>
+            <configuration>
+              <configLocation>checkstyle.xml</configLocation>
+              <encoding>UTF-8</encoding>
+              <consoleOutput>true</consoleOutput>
+            </configuration>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
     <pluginManagement>
       <plugins>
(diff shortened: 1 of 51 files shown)


pom.xml
```diff
@@ -426,6 +426,32 @@
           <serverId>ossrh</serverId>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>${version.checkstyle.plugin}</version>
```
gastaldi: Where is `version.checkstyle.plugin` declared?

pom.xml
```diff
@@ -426,6 +426,32 @@
...
+        <dependencies>
+          <dependency>
+            <groupId>com.puppycrawl.tools</groupId>
+            <artifactId>checkstyle</artifactId>
+            <version>${version.checkstyle}</version>
```
gastaldi: Where is `version.checkstyle` delcared?

pom.xml
ia3andy: in jboss-parent

gastaldi: Ah it's declared in the parent pom. You don't need to declare <version> or the dependency to com.puppycrawl.tools

pom.xml
ia3andy: it is declared only as a property not as a pluginDependency so it's not working without it.

ia3andy: I took this code from there: https://github.com/wildfly-swarm/wildfly-swarm/blob/fb119fbdd3ef22e12d7a3f37df12400407f03d1c/pom.xml#L136

gastaldi:
[test]

PR #171, opened 2018-03-09
feat(booster-catalog): Using 'latest' as the git_ref will resolve the latest release in github

gastaldi:


quintesse:
Nice!

The only thing I'd comment on is: why not use `latest`? It seems more widely used as the typical special value for specifying the latest release of something (see Docker for example)

gastaldi:
I thought about that and was undecided, because I didn't want to confuse releases with tags, but in the end it's all the same, so I'll change to `latest` anyway :)

PR #172, opened 2018-03-09
New catalog layout

quintesse:


quintesse:
Test should fail because the catalog hasn't been switched over yet

addons/launcher-addon/src/test/resources/fetch-booster-bundles.sh
```diff
@@ -6,15 +6,11 @@ mkdir -p ${TEST_DIR} > /dev/null 2>&1
 BOOSTER_TMP_DIR="${TEST_DIR}/booster-catalog"
 git clone git@github.com:fabric8-launcher/launcher-booster-catalog.git ${BOOSTER_TMP_DIR}
 cd ${BOOSTER_TMP_DIR}
-for branch in `git branch -a | grep remotes | grep -v HEAD | grep -v master `; do
-   git branch --track ${branch#remotes/origin/} $branch
```
gastaldi: Won't this command change to the correct git branch?

addons/legacy-osio-addon/src/main/java/io/fabric8/forge/generator/quickstart/ValidBoosters.java
```diff
@@ -1,68 +0,0 @@
...
-import java.util.HashSet;
-import java.util.Set;
-
-import io.fabric8.launcher.booster.catalog.Booster;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-/**
- */
-public class ValidBoosters {
```
gastaldi: YES! Goodbye headache! :+1:

addons/launcher-addon/src/test/resources/fetch-booster-bundles.sh
quintesse: There is no "correct" git branch anymore, we only have "master" nowadays for the catalog, don't we?

gastaldi: Ah, I thought those were for the boosters. In this case, nvm

PR #173, opened 2018-03-10
add stage name and description

animuk:


gastaldi:
Use the same YAML Parser API to read this information. You don't need to use any other API or any other ugly hack 😉

animuk:
Hi @gastaldi : Sure. I can use YAMP Parser API to load things. But, Could you please help me to understand the shortcoming of using the JSON way of parsing. Will appreciate on that front.

gastaldi:
Hi @animuk. 

As some people say, there's more than one way to skin a cat, but the problem is that you're introducing a hard dependency on a library that although it produces the expected result, it is not meant to be used in this context. 

See, you're parsing a YAML file, so do you think it would make sense to handle the input as a JSON content? Or do you rather rely on a library that was created just for parsing YAML content? 

Best Regards.

quintesse:
> Or do you rather rely on a library that was created just for parsing YAML content?

Not only that, but i't of course much more sensible to use a dependency that is _already in use_ by the application than adding a new one that does basically the same.

PR #176, opened 2018-03-12
fix(OSIO): Handle launch without GitOrganization specified

ia3andy:
Fixes #131


ia3andy:
I did not add tests as part of this PR because I did not find any test set regarding the osio GitSteps or the OsioEndpoint.

PR #187, opened 2018-03-14
feat(git): Add search repository by name to git api

### LINKED ISSUE #174: Give a way to filter out repositories given a string parameter
It would look like this:
```
// In repositories belonging to the logged user
searchRepositories(GitRepositoryFilter filter)

// In repositories belonging to the specified organization
searchRepositories(GitOrganization organization, GitRepositoryFilter filter)

GitRepositoryFilter {
    +withNameContaining
}
```

ia3andy:
Fixes #174


quintesse:
@ia3andy any reason why you didn't look at my suggestions on implementing this?

ia3andy:
>  any reason why you didn't look at my suggestions on implementing this?

I did look at it and answered there :)
-> GH, GL and BB apis all have a search endpoint. It is quicker than retrieving the whole content then filtering.

If it was not the case the java stream api would have been plenty enough to filter I think.

ia3andy:
[tests]

ia3andy:
[test]

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Value.Immutable(singleton = true)
+@JsonSerialize(as = ImmutableGitRepositoryFilter.class)
+@JsonDeserialize(as = ImmutableGitRepositoryFilter.class)
+public interface GitRepositoryFilter {
+
+    @Nullable String withNameContaining();
+
+    @Value.Parameter
+    @Nullable GitOrganization withOrganization();
```
gastaldi: I think it would be easier to make them return `Optional` objects to avoid the boilerplate of "if (.. == null)", but that's fine

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
```diff
@@ -0,0 +1,19 @@
...
+import org.immutables.value.Value;
+
+@Value.Immutable(singleton = true)
+@JsonSerialize(as = ImmutableGitRepositoryFilter.class)
+@JsonDeserialize(as = ImmutableGitRepositoryFilter.class)
+public interface GitRepositoryFilter {
+
+    @Nullable String withNameContaining();
+
+    @Value.Parameter
```
gastaldi: Just curious, wouldn't adding a `@Parameter` here making it part of the `ImmutableGitRepositoryFilter.of` signature?

web/src/main/java/io/fabric8/launcher/web/endpoints/GitEndpoint.java
```diff
@@ -72,7 +71,7 @@ public GitUser getUser() {
     @Produces(MediaType.APPLICATION_JSON)
     public Set<String> getRepositories(@QueryParam("organization") String organization) {
         ImmutableGitOrganization org = organization != null ? ImmutableGitOrganization.of(organization) : null;
-        return gitService.getRepositories(org).stream()
+        return gitService.getRepositories(ImmutableGitRepositoryFilter.of(org)).stream()
```
gastaldi: Is the org always necessary for the creation of ImmutableGitRepositoryFilter? See my comment about @Parameter

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
ia3andy: yes (I wanted it as a shortcut since it's a common usage of this filter), and the singleton makes a of() (to build an empty filter). Do you see a reason why I shouldn't use it?

web/src/main/java/io/fabric8/launcher/web/endpoints/GitEndpoint.java
ia3andy: Ok the singleton should answer you question.

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
gastaldi: You say it is a @Parameter but at the same time it is @Nullable, so it is confusing. @Parameters shouldn't be @Nullable afaik

gastaldi: You say it is a @Parameter but at the same time it is @Nullable, so it is confusing. @Parameters shouldn't be @Nullable afaik

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
ia3andy: We could talk forever about this one :)
At the end, I think, the most important is to have a common approach on how we use `Optional`. I don't mind if you want me to change it..

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
ia3andy: I don't see any reason why a parameter shouldn't be @nullable, but let's use `Optional` if you would feel more comfy then lol

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
gastaldi: Not asking to Change it, just an idea ;)

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/api/GitRepositoryFilter.java
quintesse: I don't agree. According to best practices `Optional` should _not_ be used as parameters. And `@Nullable` parameters are perfectly normal

gastaldi: @quintesse I may be wrong, but I don't think using Optional here will create it as a method parameter

ia3andy: Using Optional and @parameter end up to this:
```Java
/**
   * Construct a new immutable {@code GitRepositoryFilter} instance.
   * @param withOrganization The value for the {@code withOrganization} attribute
   * @return An immutable GitRepositoryFilter instance
   */
  public static ImmutableGitRepositoryFilter of(Optional<GitOrganization> withOrganization) {
    return validate(new ImmutableGitRepositoryFilter(withOrganization));
  }

```

gastaldi: Horrible. I thought that would work only for method returns

ia3andy:
At the end I did change to: 

- kept @nullable because the Immutables is using Optional as param
- removed the `@parameter` and just used the builder (which is clearer at the end) `ImmutableGitRepositoryFilter.builder().withOrganization(getTestOrganization()).build()`
- kept the singleton=true to have `of()` when you want both attribute `null`

gastaldi:
The CI builds are failing because the credentials are wrong. Asked the SD team to fix that: https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1468

web/src/main/java/io/fabric8/launcher/web/endpoints/GitEndpoint.java
```diff
@@ -71,8 +71,10 @@ public GitUser getUser() {
     @Secured
     @Produces(MediaType.APPLICATION_JSON)
     public Set<String> getRepositories(@QueryParam("organization") String organization) {
```
gastaldi: I think you can replace the `String organization` with `ImmutableGitOrganization organization`?

gastaldi: Or `GitOrganization organization` even, not sure if the Jackson serialization can handle that

gastaldi:
Fixed in 3322bf284710731e75dd44fba8c46675548f7b21

PR #192, opened 2018-03-14
Catalog metadata

quintesse:
This adds `metadata` sections to the REST results for missions, runtimes and versions. THis should be a flexible enough solution for https://github.com/fabric8-launcher/launcher-planning/issues/120 and https://github.com/fabric8-launcher/launcher-planning/issues/121


base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        return builder;
+    }
+
+    /**
+     * Converts any <code>Iterable</code> of type <code>&lt;Object&gt;</code>
+     * to a <code>JsonArray</code>. The items of the array can itself be objects,
+     * arrays or simple values
+     */
+    @SuppressWarnings("unchecked")
+    public static JsonArrayBuilder listToJson(Iterable<Object> list) {
```
gastaldi: Technically the parameter is not a `List`, but an `Iterable` :) 

Perhaps just name both methods `toJson`?

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
```diff
@@ -53,11 +53,14 @@ public Response getMissions() {
...
             JsonObjectBuilder mission = createObjectBuilder()
                     .add("id", m.getId())
-                    .add("name", m.getName())
-                    .add("suggested", m.isSuggested());
+                    .add("name", m.getName());
             if (m.getDescription() != null) {
                 mission.add("description", m.getDescription());
             }
+            if (m.getMetadata() != null && !m.getMetadata().isEmpty()) {
+                mission.add("metadata", mapToJson(m.getMetadata()));
```
gastaldi: Can we make it part of the same object? I think adding `metadata` here would confuse the consumer

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
gastaldi: Also, a test class would be nice

gastaldi: One solution would be to pass the JsonArrayBuilder as a parameter here

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
```diff
@@ -0,0 +1,84 @@
...
+            } else if (entry.getValue() instanceof Long) {
+                builder.add(entry.getKey(), (Long)entry.getValue());
+            } else if (entry.getValue() instanceof Integer) {
+                builder.add(entry.getKey(), (Integer)entry.getValue());
+            } else if (entry.getValue() instanceof BigInteger) {
+                builder.add(entry.getKey(), (BigInteger)entry.getValue());
+            } else if (entry.getValue() instanceof BigDecimal) {
+                builder.add(entry.getKey(), (BigDecimal) entry.getValue());
+            } else {
+                builder.add(entry.getKey(), entry.getValue().toString());
```
gastaldi: What if getValue() returns null?

gastaldi:
The CI builds are failing because the credentials are wrong. Asked the SD team to fix that: https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/1468

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
quintesse: Confuse? How?

And I don't like the idea of metadata elements possibly overwriting elements in the parent object. That's something that would not be clear at all to the person adding the metadata in the .yaml file.

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
quintesse: I had that before, but I didn't really like it. To me `toXxxx()` methods should be part of an actual object (like `path.toFile()` or `obj.toString()`), not as static methods. You lose a bit of context that way I think. I can change the `Iterable` to `List` though ;-)

gastaldi: `Iterable` is fine, it's just the method name that may need to be changed.

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
gastaldi: That is a good point, in that case we need to figure out how to describe that in the Swagger document.

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
gastaldi: But I think passing the object as a parameter would be much better, what if you want to add that result to an existing builder?

quintesse: That seems like a strange thing to do. Do you have a use-case for that? And even so I'd say that it's easy enough to add overloads adding that extra parameter if at some future date we need that, right?

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
quintesse: Perhaps a type of `object` and an explanation that it is a copy of what is found inside the booster catalog itself?

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
quintesse: Look ☝️

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
gastaldi: I don't think it's strange, but I can live with that 👍

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
gastaldi: Ah cool, I missed that in my first look :+1:

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
gastaldi: WFM

gastaldi:
Approved, but please add a test case to the class in the base module

gastaldi:
Retest this please

gastaldi:
[test]

gastaldi:
[test]

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
gastaldi: I think it would be nice to move the `if (entry.getValue() == null) ... addNull` as the first condition

gastaldi:
Looks like I screwed up the commit list while resolving the conflicts using the Github UI 😛

PR #200, opened 2018-03-17
Adding proper validation to the space path

ravsa:
Space path name requires validation:

- Application name must contain only letters (uppercase and lowercase), numbers, hyphens, underscores.
- It can not start and end with hyphens, spaces or underscores.
- It should be minimum 4 characters long.
- It should be maximum 63 characters long.
- It should start with a forward slash.

### DIFF
--- a/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/projectiles/OsioImportProjectileContext.java
+++ b/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/projectiles/OsioImportProjectileContext.java
@@ -3,6 +3,7 @@ package io.fabric8.launcher.osio.projectiles;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
 import javax.ws.rs.FormParam;
+import javax.validation.constraints.Size;
 
 public class OsioImportProjectileContext {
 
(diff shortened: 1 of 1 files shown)


gastaldi:
Thanks

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/projectiles/OsioImportProjectileContext.java
```diff
@@ -23,7 +24,11 @@
 
     @FormParam("spacePath")
     @NotNull
-    @Pattern(message = "Space Path should start with a /", regexp = "\\/[a-zA-Z]*")
+    @Size(message = "Space Path must be in the range of 4 to 63 characters long", min = 5, max = 63)
```
quintesse: Text says 4-63 but code says 5-63

gastaldi: I think it is considering the initial slash?

quintesse: Well then the code should say 5-64, shouldn't it?

gastaldi: Ah good catch, yeah, I agree

gastaldi: @ravsa what do you think? If you also agree, can you provide another PR with this fix? Thanks a million!

ravsa: nice catch, @quintesse 
yes, I will raise another PR to fix it.

PR #203, opened 2018-03-18
feat(osio): insert dependencies into pom.xml

ravsa:
API end-point `/osio/launch` will have an optional field `dependency` which will take dependencies in the format of `dependency=groupId:artifactId:version&dependency=groupId:artifactId:version` from the dependency editor to insert the newly added dependencies into the pom.xml file of selected boosters.

Closes: https://github.com/openshiftio/openshift.io/issues/2644


gastaldi:
@ravsa can you create the commit message using the http://karma-runner.github.io/2.0/dev/git-commit-msg.html pattern?

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,80 @@
...
+public class InsertMavenDependenciesPreparer implements ProjectilePreparer {
+    @Inject
+    private ResourceFactory resourceFactory;
+
+    @Override
+    public void prepare(Path projectPath, RhoarBooster booster, ProjectileContext context) {
+        if (!(context instanceof CreateProjectileContext)) {
+            return;
+        }
+        Logger LOG = Logger.getLogger(InsertMavenDependenciesPreparer.class.getName());
```
gastaldi: This can be a static final attribute

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/projectiles/OsioProjectileContext.java
```diff
@@ -35,6 +35,10 @@
     @DefaultValue("1.0.0")
     private String projectVersion;
 
+    @FormParam("dependencySnapshot")
+    @DefaultValue("{}")
+    private String dependencySnapshot;
```
gastaldi: Using String to represent this field may be harder to validate and parse the content. Ideally it would be nice to create Immutable VOs(using @Immutable) to define what's expected in this attribute

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,80 @@
...
+        CreateProjectileContext createProjectileContext = (CreateProjectileContext) context;
+        DirectoryResource projectDirectory = resourceFactory.create(projectPath.toFile()).as(DirectoryResource.class);
+        MavenModelResource modelResource = projectDirectory.getChildOfType(MavenModelResource.class, "pom.xml");
+
+        // Perform model changes
+        if (modelResource.exists()) {
+            Model model = modelResource.getCurrentModel();
+            String dependencySnapshot = createProjectileContext.getDependencySnapshot();
+            JsonReader reader = Json.createReader(new StringReader(dependencySnapshot));
```
gastaldi: Put in the try-finally block instead. Then you don't need to explicitly close it

gastaldi:
Some suggestions to make it more intuitive:

- rename `dependencySnapshot` to `dependencies`
- rename `ecosystem` to `platform` (we use this terminology in the Jenkins pipelines)
- rename `package` to `coordinate` (maven uses this nomenclature)

What is request ID for?

gastaldi:
Perhaps the only information needed is `dependency` which may be passed in the request as query params, like `dependency=a&dependency=b`.

Try to keep it simple

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/projectiles/OsioProjectileContext.java
gastaldi: If you pass only the dependencies, this can be a `String[]` containing only the dependencies in the `groupId:artifactId:version` format

core/core-api/src/main/java/io/fabric8/launcher/core/api/CreateProjectileContext.java
```diff
@@ -19,4 +19,6 @@
     String getArtifactId();
 
     String getProjectVersion();
+
+    String getDependencySnapshot();
```
gastaldi: This is specific to OSIO, remove from core

addons/launcher-addon/src/main/java/io/fabric8/launcher/addon/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,80 @@
...
+import io.fabric8.launcher.core.spi.ProjectilePreparer;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.Dependency;
+import org.jboss.forge.addon.maven.resources.MavenModelResource;
+import org.jboss.forge.addon.resource.DirectoryResource;
+import org.jboss.forge.addon.resource.ResourceFactory;
+
+
+@ApplicationScoped
+public class InsertMavenDependenciesPreparer implements ProjectilePreparer {
```
gastaldi: This is specific to OSIO, put in the OSIO addon

web/src/main/java/io/fabric8/launcher/web/endpoints/inputs/LaunchProjectileInput.java
```diff
@@ -51,6 +51,10 @@
     @DefaultValue("0")
     private String step;
 
+    @FormParam("dependencySnapshot")
+    @DefaultValue("{}")
+    private String dependencySnapshot;
```
gastaldi: Specific to OSIO, don't add to core or web

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,54 @@
...
+            try {
+                Model model = Maven.readModel(pom);
+                // Get the list of dependencies in the format of 'groupId:artifactId:version'
+                for (String dep : dependencies) {
+                    Dependency dependency = new Dependency();
+                    String[] groupIdArtifactIdVersion = dep.split(":");
+                    dependency.setGroupId(groupIdArtifactIdVersion[0]);
+                    dependency.setArtifactId(groupIdArtifactIdVersion[1]);
+                    dependency.setVersion(groupIdArtifactIdVersion[2]);
+                    model.addDependency(dependency);
```
gastaldi: What if the dependency already exists? Maven will throw an error if duplicate dependencies are found in the pom.xml

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,54 @@
...
+import io.fabric8.launcher.osio.projectiles.context.OsioProjectileContext;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.Dependency;
+
+import javax.enterprise.context.RequestScoped;
+
+@RequestScoped
+public class InsertMavenDependenciesPreparer implements ProjectilePreparer {
+
+    private final Logger LOG = Logger.getLogger(InsertMavenDependenciesPreparer.class.getName());
```
gastaldi: Use `logger`  (in lowercase)

gastaldi: And make it static

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,54 @@
...
+import io.fabric8.launcher.booster.catalog.rhoar.RhoarBooster;
+import io.fabric8.launcher.core.api.ProjectileContext;
+import io.fabric8.launcher.core.spi.ProjectilePreparer;
+import io.fabric8.launcher.osio.projectiles.context.OsioProjectileContext;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.Dependency;
+
+import javax.enterprise.context.RequestScoped;
+
+@RequestScoped
```
gastaldi: Why RequestScoped? It doesn't seem to store any state

addons/osio-addon/src/test/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparerTest.java
```diff
@@ -0,0 +1,124 @@
...
+                BoosterPredicates.withMission(new Mission("rest-http"))
+                        .and(BoosterPredicates.withRuntime(new Runtime("vert.x"))
+                                .and(BoosterPredicates.withVersion(new Version("redhat")))));
+
+
+        final RhoarBooster rhoarBooster = booster.get();
+        final Path rootPath = temporaryFolder.getRoot().toPath();
+        boosterCatalogService.copy(rhoarBooster, rootPath);
+
+        final InsertMavenDependenciesPreparer insertMavenDependenciesPreparer = new InsertMavenDependenciesPreparer();
```
gastaldi: Although that is fine, it would be much easier if you just created a directory with a known pom.xml and passed null to the booster param (don't forget to mark it as @Nullable)

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
ravsa: @gastaldi what should be the expected behavior if a user wants to add a dependency which already exists but the version is different.

gastaldi: Use the version that is in the list of dependencies in the ProjectileContext

ravsa: ok, Thanks

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,70 @@
...
+                log.log(Level.SEVERE, "An exception occurred while adding the dependencies. " + e.toString());
+            }
+
+        }
+    }
+
+    public boolean isDuplicateDependency(List<String> allDependencies, String dependency) {
+        try {
+            String[] groupIdArtifactIdVersion = dependency.split(":");
+            return allDependencies.contains(groupIdArtifactIdVersion[0] + ":" + groupIdArtifactIdVersion[1]);
```
gastaldi: This will return false if the version is sent in the list of dependencies

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,70 @@
...
+        } catch (Exception e) {
+            log.log(Level.WARNING, "An exception occurred while checking for duplicate dependency. " + e.toString());
```
gastaldi: I don't think this code throws any Exception

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/projectiles/context/OsioProjectileContext.java
```diff
@@ -37,6 +38,9 @@
     @DefaultValue("1.0.0")
     private String projectVersion;
 
+    @FormParam("dependency")
+    private List<String> dependencies;
```
gastaldi: I think it would be better if you changed this to `List<Dependency> dependencies` and used a `ParamConverter` to convert the String to Dependency. Here is an example: https://github.com/fabric8-launcher/launcher-backend/commit/aedbebb8da169abb99e7bb38ccaf4bc3bb9a7cce

gastaldi: Here is an explanation on how ParamConverters work: https://abhirockzz.wordpress.com/2016/01/20/using-parameter-converters-in-jax-rs/

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
ravsa: yes, It will return false but here I'm sending list of dependencies in the format `[ "groupId:artifactId", "groupId:artifactid"]`

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
ravsa: I added try-catch block so that it can handle exceptions like `ArrayIndexOutOfBoundsException` if the dependency is not in the right format.

gastaldi: Don't do that, always validate and parse your input adequately. Making it a `List<Dependency>` would ensure that you have the correct format and validate if the input is correct right in the beginning

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/providers/DependencyParamConverter.java
```diff
@@ -0,0 +1,42 @@
...
+import static javax.json.Json.createObjectBuilder;
+
+@ApplicationScoped
+public class DependencyParamConverter implements ParamConverter<Dependency> {
+
+    @Override
+    public Dependency fromString(final String dep) {
+        Dependency dependency = new Dependency();
+        try {
+            String[] groupIdArtifactIdVersion = dep.split(":");
```
gastaldi: I really think you should validate if the input is valid instead of letting throw an ArrayIndexOutOfBoundsException here.

addons/osio-addon/src/test/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparerTest.java
gastaldi: Actually I take that back. Do you mind simplifying the test a bit? That looks unnecessarily complex now that I see it again

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/providers/DependencyParamConverter.java
```diff
@@ -0,0 +1,42 @@
...
+            dependency.setGroupId(groupIdArtifactIdVersion[0]);
+            dependency.setArtifactId(groupIdArtifactIdVersion[1]);
+            dependency.setVersion(groupIdArtifactIdVersion[2]);
```
gastaldi: Version is required?

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/providers/DependencyParamConverterProvider.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.ws.rs.ext.ParamConverterProvider;
+import javax.ws.rs.ext.Provider;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@ApplicationScoped
+public class DependencyParamConverterProvider implements ParamConverterProvider {
+    @Inject
+    private Instance<DependencyParamConverter> dependencyParamConverter;
```
gastaldi: You can @Inject DependencyParamConverter directly here, since they share the same scope

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,59 @@
...
+                ).collect(Collectors.toList());
+
+                for (Dependency dep : dependencies) {
+                    if (!isDuplicateDependency(boosterDependencies, dep)) {
+                        model.addDependency(dep);
+                    }
+                }
+                Maven.writeModel(model);
+            } catch (Exception e) {
+                log.log(Level.SEVERE, "An exception occurred while adding the dependencies. " + e.toString());
```
gastaldi: Pass the original exception here

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/preparers/InsertMavenDependenciesPreparer.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class InsertMavenDependenciesPreparer implements ProjectilePreparer {
+
+    private static final Logger log = Logger.getLogger(InsertMavenDependenciesPreparer.class.getName());
+
+
+    @Override
+    public void prepare(Path projectPath, RhoarBooster booster, ProjectileContext context) {
```
gastaldi: Add a @Nullable in the RhoarBooster param

PR #205, opened 2018-03-20
refactor(service): remove spi interface

ia3andy:


ia3andy:
[test]

quintesse:
@ia3andy I think it's a good idea to put a short description as to _why_ you want to do this. We were there doing the discussion this morning but @gastaldi won't have any idea when he sees this. 

You should probably have made it part of the commit message itself. Because this is a big enough change that the "karma commit" style of a 1 line title + explanation is called for.

gastaldi:
I read the conversation in Slack, but I am not 100% sure that we should be getting rid of the SPI interfaces. I think an easy solution is to change the @Produces method to return the SPI instead, since they specialize the original service (haven't tested yet)?

ia3andy:
@quintesse you are absolutely right, my bad, I have been a bit lazy..

Even if the problem that raised this PR was the fact it was not possible to cast an injected proxy into one of his implementing interface (spi in this case). The justifications for this PR are different:
- It does make sense when you can create to be able to delete as well in a service and if we move the delete methods in the api, there is nothing left in the spi
- The token already takes care about the security concerns
- @gastaldi producing the spi would be really close to having no spi
- It just more simple

ia3andy:
I made the PR because it was 5 min, and it's a proposal. It really doesn't have to be merged if we think it's not better with it :)

gastaldi:
@ia3andy I see your point and indeed it is simpler. I think there is a concern in the public API about what it *can* do versus what it *should* do. The *delete* method is clearly not something the public API should do, but invoking it in an environment where the token disallows it will block this operation. @quintesse @edewit what are your thoughts about it?

edewit:
I like that our API only has methods that are used for the actual work that needs to be done to launch an application, so adding these methods to the public API when their only used in tests is not a solution IMO. Also our ["github app"](https://github.com/settings/connections/applications/aa087c110f8e83f6e1c1) should have just enough rights to do it's work and that means it doesn't have delete rights.
:-1: from me, I think the better solution would be to use a username and token to get the GitServiceSpi

ia3andy:
@edewit I don't disagree :)
Let throw this PR away...

PR #208, opened 2018-03-20
feat(cors): Replaced CorsFilter with ebay library

gastaldi:
Artifacts outside the JAX-RS domain (/swagger.yaml for example) may also need CORS headers


gastaldi:
This is the library used: https://github.com/eBay/cors-filter

PR #214, opened 2018-03-21
feat(osio): Creation of codebase using the WIT service

### LINKED ISSUE #209: OSIO: Call WIT's codebase/create after importing a repository or launching a booster
Following the discussion over https://github.com/fabric8-services/fabric8-wit/pull/1925#issuecomment-374505275, the launcher-backend should invoke the WIT API to create the codebase.

`/osio/import` should also validate if the build config already exists (see https://github.com/fabric8-services/fabric8-wit/pull/1925#issuecomment-374959071)

Endpoint documentation: http://swagger.goa.design/?url=github.com%2Ffabric8-services%2Ffabric8-wit%2Fdesign#!/space_codebases/space_codebases_create

/cc @corinnekrych @aslakknutsen

### LINKED ISSUE #213: /osio/launch and /osio/import should receive the spaceId
Instead of receiving the `spacePath`, the backend should receive the `spaceId` and perform the necessary lookup to get the space path. This is because the ID is needed to invoke other services

Needed for #209

gastaldi:
Fixes #213 
Fixes #209

### DIFF
--- a/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/WitSteps.java
+++ b/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/WitSteps.java
@@ -1,0 +1,74 @@
+package io.fabric8.launcher.osio.steps;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.fabric8.launcher.base.http.ExternalRequest;
+import io.fabric8.launcher.base.http.HttpException;
+import io.fabric8.launcher.osio.EnvironmentVariables;
+import io.fabric8.launcher.osio.tenant.Tenant;
+import io.fabric8.launcher.service.git.api.GitRepository;
+import okhttp3.MediaType;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+
+/**
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
+ */
+@RequestScoped
+public class WitSteps {
+
+    @Inject
+    private Tenant tenant;
+
+    private static final Logger logger = Logger.getLogger(WitSteps.class.getName());
+
+    public void createCodebase(String spaceId, String stackId, GitRepository repository) {
+        String payload = "{\n" +
+                "  \"data\": {\n" +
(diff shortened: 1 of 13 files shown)


addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/WitSteps.java
```diff
@@ -0,0 +1,74 @@
...
+                    StringWriter sw = new StringWriter();
+                    PrintWriter pw = new PrintWriter(sw, true);
+                    for (JsonNode error : errors) {
+                        pw.println(error.get("detail").asText());
+                    }
+                    message = sw.toString();
+                } catch (IOException e) {
+                    logger.log(Level.WARNING, "Error while reading error from WIT", e);
+                }
+                throw new HttpException(response.code(), message);
```
aslakknutsen: Are you failing the whole step on 409 Conflict? e.g. it already exists in Codebases API

This can be ignored for now as there is no connection in the 'beginning' of the wizard to verify what is in the codebase API, as talked about here https://github.com/fabric8-services/fabric8-wit/pull/1925#issuecomment-374982046

gastaldi: I didn't see 409 in the [Swagger document](http://swagger.goa.design/?url=github.com%2Ffabric8-services%2Ffabric8-wit%2Fdesign#!/space_codebases/space_codebases_create)

aslakknutsen: hehe 

It's the topic of the PR where this whole discussion started:
https://github.com/fabric8-services/fabric8-wit/pull/1925

The question was; Are we ready to handle that case. Now you moved the call to the backend of launcher and don't handle it, we're waiting for UI to remove their call and send SpaceID, then we can move forward 1925

gastaldi: The new UI is the only one calling this endpoint, so it shouldn't affect the legacy wizard

gastaldi: That said, I don't think they call the WIT endpoint after this endpoint in the new UI but I may be mistaken

aslakknutsen: While probably all correct, not really related :)

Your code would still fail if Codebase API returns 409.

gastaldi: Fixed in https://github.com/fabric8-launcher/launcher-backend/commit/2b20a52c17aaa6d2e901ce36c20ab3a8c98931b5

PR #216, opened 2018-03-23
feature(config-map): Added way to inject jenkins file snippets

edewit:
find `// INJECT FRAGMENT` and replaces that with found `Jenkinsfile.snippets` in `.openshiftio` folder

### DIFF
--- a/addons/osio-addon/src/test/resources/jenkinsfile-preparer-test/pipeline/inject/Jenkinsfile
+++ b/addons/osio-addon/src/test/resources/jenkinsfile-preparer-test/pipeline/inject/Jenkinsfile
@@ -1,0 +1,10 @@
+#!/usr/bin/groovy
+@Library('github.com/fabric8io/fabric8-pipeline-library@master')
+def canaryVersion = "1.0.${env.BUILD_NUMBER}"
+
+// INJECT FRAGMENT setup
+
+mavenNode {
+  checkout scm
+  mavenCI{}
+}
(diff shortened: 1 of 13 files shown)


addons/osio-addon/src/test/resources/jenkinsfile-preparer-test/pipeline/inject/Jenkinsfile
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/groovy
+@Library('github.com/fabric8io/fabric8-pipeline-library@master')
+def canaryVersion = "1.0.${env.BUILD_NUMBER}"
+
+// INJECT FRAGMENT
```
gastaldi: The fragment needs a name and the name is the same that `Jenkinsfile.<name>.fragment` uses

gastaldi: Although I think we could call `SNIPPET` instead of fragment

edewit: on my fragment inject PR you didn't want to call it `config-map` https://github.com/fabric8io/fabric8-jenkinsfile-library/pull/28

gastaldi: Yes, I wanted it to be called `pre-deploy` or something like that (like a hook name)

PR #226, opened 2018-03-27
fix(osio): build stuck pending due to invalid job-path

edewit:
fixes https://github.com/fabric8-launcher/launcher-planning/issues/139


addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/OpenShiftSteps.java
```diff
@@ -115,15 +115,15 @@ private BuildConfig createBuildConfigObject(OsioProjectile projectile, GitReposi
         String gitUrl = repository.getGitCloneUri().toString();
         BuildConfig buildConfig = Builds.createDefaultBuildConfig(projectile.getOpenShiftProjectName(), gitUrl, getJenkinsUrl());
         Map<String, String> currentAnnotations = KubernetesHelper.getOrCreateAnnotations(buildConfig);
-        currentAnnotations.putAll(getBuildConfigAnnotations(projectile));
+        currentAnnotations.putAll(getBuildConfigAnnotations(projectile, repository.getFullName()));
         return buildConfig;
     }
 
-    private Map<String, String> getBuildConfigAnnotations(OsioProjectile projectile) {
+    private Map<String, String> getBuildConfigAnnotations(OsioProjectile projectile, String gitRepositoryName) {
```
gastaldi: It would be better if the `GitRepository` object was used in this parameter. Typesafe ftw! ;)

edewit: really if you only need one thing from that object why pass in the entire object

gastaldi: Typesafe. A String doesn't have any meaning. A GitRepository does

gastaldi: Why are you worried about passing the object? It's not a remote call

PR #232, opened 2018-03-28
refactor(identity): Change IdentityProvider signature

### LINKED ISSUE #230: Change IdentityProvider signature and make it @RequestScoped
- Remove the token from the signature: Optional<Identity> getIdentity(String service);
- Read the Authorization header token when creating the IdentityProvider for the request scope and inject it.
- Sometimes "token" means only the token and sometime authorization header value (with a prefix like Bearer). It is a bit confusing and could lead to mistake. -> Try to make it more uniform
- Remove all call that directly parse the authorization header (instead call the new request scoped IdentityProvider which already contain it)

ia3andy:
and make it @RequestScoped

Fixes #230

### DIFF
--- a/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
+++ b/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
@@ -2,37 +2,38 @@ package io.fabric8.launcher.osio.identity;
 
 import java.util.Optional;
 
-import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
 
 import io.fabric8.launcher.base.identity.Identity;
 import io.fabric8.launcher.base.identity.IdentityFactory;
-import io.fabric8.launcher.base.identity.TokenIdentity;
 import io.fabric8.launcher.core.spi.Application;
 import io.fabric8.launcher.core.spi.IdentityProvider;
 import io.fabric8.launcher.osio.client.OsioApiClient;
-import io.fabric8.launcher.osio.client.OsioApiClientImpl;
+import io.fabric8.launcher.osio.client.Tenant;
 
-import static io.fabric8.launcher.base.identity.IdentityFactory.createFromToken;
-import static io.fabric8.launcher.base.identity.IdentityHelper.removeBearerPrefix;
 import static io.fabric8.launcher.core.spi.Application.ApplicationType.OSIO;
 import static io.fabric8.launcher.osio.OsioConfigs.ExternalServices.getGithubServiceName;
 
 
-@ApplicationScoped
+@Dependent
 @Application(OSIO)
 public class OsioIdentityProvider implements IdentityProvider {
 
+    @Inject
+    private OsioApiClient osioApiClient;
+
+    @Inject
+    private Tenant tenant;
 
     @Override
-    public Optional<Identity> getIdentity(final String service, final String authorizationHeader) {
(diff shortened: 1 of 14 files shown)


addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
```diff
@@ -2,37 +2,38 @@
...
-        final TokenIdentity osioIdentity = createFromToken(removeBearerPrefix(authorizationHeader));
-        final OsioApiClient osioApiClient = new OsioApiClientImpl(osioIdentity);
+    public Optional<Identity> getIdentity(final String service) {
         switch (service) {
             case IdentityProvider.ServiceType.GITHUB:
                 return osioApiClient.getTokenForService(getGithubServiceName())
                         .map(IdentityFactory::createFromToken);
             case IdentityProvider.ServiceType.OPENSHIFT:
-                return Optional.of(osioIdentity);
+                return Optional.of(osioApiClient.getTenant().getIdentity());
```
gastaldi: Why not return tenant.getIdentity() here?

web/src/main/java/io/fabric8/launcher/web/api/OpenShiftResource.java
```diff
@@ -89,11 +91,12 @@ public JsonArray getSupportedOpenShiftClusters(@HeaderParam(HttpHeaders.AUTHORIZ
                     .forEach(arrayBuilder::add);
         } else {
             final KeycloakService keycloakService = this.keycloakServiceInstance.get();
+            final String keycloakToken = removeBearerPrefix(authorization);
```
gastaldi: Why remove the bearer to add it back in KeycloakService?

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
```diff
@@ -2,37 +2,38 @@
...
+import io.fabric8.launcher.osio.client.Tenant;
 
-import static io.fabric8.launcher.base.identity.IdentityFactory.createFromToken;
-import static io.fabric8.launcher.base.identity.IdentityHelper.removeBearerPrefix;
 import static io.fabric8.launcher.core.spi.Application.ApplicationType.OSIO;
 import static io.fabric8.launcher.osio.OsioConfigs.ExternalServices.getGithubServiceName;
 
 
-@ApplicationScoped
+@Dependent
```
gastaldi: Afaik Using @Dependent will make the bean reuse the scope of the consumer. If you are injecting it in an @ApplicationScoped endpoint (like ValidationResource) it will assume the parent scope.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
ia3andy: mistake :)

web/src/main/java/io/fabric8/launcher/web/api/OpenShiftResource.java
ia3andy: Yes you are right, the other way around makes more sense: KeycloakService.Optional<Identity> getIdentity(String provider, String authorizationHeader)

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
ia3andy: I changed it in ValidationResource by placing it in the method params ;)

ia3andy:
@gastaldi it should be alright now

PR #234, opened 2018-03-28
fix(auth): ensure that git credentials are set in jenkins

ia3andy:


ia3andy:
@edewit @gastaldi I didn't change the PR, just rebased it and moved the dependency declaration to dependencyManagement.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/jenkins/JenkinsGitCredentials.java
```diff
@@ -0,0 +1,96 @@
...
+    JenkinsGitCredentials(String authorization, String gitToken) {
+        this.authorization = authorization;
+        this.gitToken = gitToken;
+    }
+
+    private static final String GET_URL = pathJoin(OsioConfigs.getJenkinsUrl(), "/credentials/store/system/domain/_/credentials/cd-github/");
+
+    private static final String CREATE_URL = pathJoin(OsioConfigs.getJenkinsUrl(), "/credentials/store/system/domain/_/createCredentials");
+
+    private String authorization;
```
gastaldi: This field is not used, remove it

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/jenkins/JenkinsGitCredentials.java
```diff
@@ -0,0 +1,96 @@
...
+import okhttp3.RequestBody;
+
+import static io.fabric8.utils.URLUtils.pathJoin;
+
+@RequestScoped
+public class JenkinsGitCredentials {
+    public JenkinsGitCredentials() {
+    }
+
+    JenkinsGitCredentials(String authorization, String gitToken) {
```
gastaldi: I assume this constructor is using for tests only. I think it should be better if it accepted an `IdentityProvider`?

Issue #238, opened 2018-03-28
IllegalArgumentException: The given name is not a valid git repository name: https://github.com/SecurityCentral/ssptool.git.


gastaldi:
https://errortracking.prod-preview.openshift.io/openshift_io/fabric8-launcher-backend/issues/2120/

```
IllegalArgumentException: The given name is not a valid git repository name: https://github.com/SecurityCentral/ssptool.git.
    at io.fabric8.launcher.service.git.GitHelper.checkGitRepositoryNameArgument(GitHelper.java:72)
    at io.fabric8.launcher.service.git.github.KohsukeGitHubService.createRepository(KohsukeGitHubService.java:144)
    at io.fabric8.launcher.service.git.github.KohsukeGitHubService.createRepository(KohsukeGitHubService.java:185)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
...
(76 additional frame(s) were not displayed)

Error while launching project
```

We need to validate the input, probably by adding a @Pattern in the gitRepository attribute in `OsioImportProjectileContext` and `LaunchProjectileInput`

PR #239, opened 2018-03-29
fix(endpoint): Add pattern checking to git params

ia3andy:
Fixes: #238

### DIFF
--- a/addons/osio-addon/pom.xml
+++ b/addons/osio-addon/pom.xml
@@ -15,6 +15,10 @@
       <groupId>io.fabric8.launcher</groupId>
       <artifactId>launcher-core-api</artifactId>
     </dependency>
+    <dependency>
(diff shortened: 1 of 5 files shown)


addons/osio-addon/pom.xml
```diff
@@ -15,6 +15,10 @@
       <groupId>io.fabric8.launcher</groupId>
       <artifactId>launcher-core-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.fabric8.launcher</groupId>
```
gastaldi: Depend on impl? Better move the Githelper class to API (or the constant to GitService) instead

ia3andy: Ok my bad I thought it was already a a transitive dep, but it was in scope test. I will move it ;)
Go in Holiday @gastaldi  loooool

PR #240, opened 2018-03-29
fix(arquillian-preparer): insert 'app.name' property if not present in openshift extension

dipak-pawar:
*Proposed Changes*
* Add `app.name` for `openshift` extension with application `artifactId` in  `arquillian.xml` with. Previously it was updating only if `app.name` is already present.
* Fix unit test.


gastaldi:
Thanks

PR #243, opened 2018-03-30
refactory(identity): can @Inject TokenIdentity Authorization in RequestScope

ia3andy:


ia3andy:
@gastaldi I think I found a good solution:
- add toRequestHeaderKey and toRequestHeaderValue as default in Identity
- now I can make Identities package protected
- now I can remove (new) Authorization and use TokenIdentity as Authorization in IdentityProvider
- rename GitHelper to Gits, and remove execute from it (use directly ExternalRequest instead)
- make addBearerPrefix/removeBearerPrefix less flexible

drawback: I didn't manage to split it into multiple commits :/

I think it's pretty clean now and it will be difficult to go further.

What's you opinion?

gastaldi:
OsioIdentityProvider shouldn't take a HttpServletRequest in the constructor, but just the Token identity or something similar

base/src/main/java/io/fabric8/launcher/base/http/ExternalRequest.java
```diff
@@ -59,8 +60,8 @@ public static boolean execute(Request request) {
 
     public static <T> Optional<T> executeAndParseJson(Request request, Function<JsonNode, T> jsonNodeFunction) {
         try (Response response = client.newCall(request).execute()) {
+            final ResponseBody body = response.body();
```
gastaldi: I think this may fail if response isn't successful, but I may be wrong

base/src/test/java/io/fabric8/launcher/base/identity/IdentityFactoryTest.java
```diff
@@ -4,6 +4,9 @@
 import org.junit.Rule;
```
gastaldi: Probably this class name needs to change since IdentityFactory is gone

base/src/main/java/io/fabric8/launcher/base/http/ExternalRequest.java
ia3andy: I took it from the GitHelper and tested it there. Since we never too careful https://github.com/fabric8-launcher/launcher-backend/issues/247

base/src/test/java/io/fabric8/launcher/base/identity/IdentityFactoryTest.java
ia3andy: right !!

quintesse:
Btw, I have only one comment for this PR: you _really_ should improve your commit messages :)

The first one "can Inject TokenIdentity Authorization in RequestScope" is just very meager information for such a huge commit. I look at that and I have no idea of all the things that you've done there, except if I go and read all the code very carefully. IMO you should put enough information in your commit message that I can understand what you did _before_ looking at the code.

PR #245, opened 2018-04-02
refactor(core): Introduced Projectile.getEventConsumer

gastaldi:
Ease testing and may be a first step to support server side events in the future

### DIFF
--- a/base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
+++ b/base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
@@ -16,7 +16,7 @@ import static javax.json.Json.createObjectBuilder;
 
 public final class JsonUtils {
 
-    private static final ObjectMapper mapper = new ObjectMapper();
+    private static final ObjectMapper MAPPER = new ObjectMapper();
 
     private JsonUtils() {
         throw new IllegalAccessError("Utility class");
@@ -24,12 +24,12 @@ public final class JsonUtils {
 
 
     public static String toString(Object obj) throws IOException {
-        return mapper.writeValueAsString(obj);
+        return MAPPER.writeValueAsString(obj);
     }
 
 
     public static JsonNode readTree(String content) throws IOException {
-        return mapper.readTree(content);
+        return MAPPER.readTree(content);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


gastaldi:
I thought of calling it "Beacon" (as in a beacon attached to a missile to know its location). Wdyt?

gastaldi:
Maybe a better idea is to change the MissionControl.launch method signature to accept a Projectile and a Consumer instead of adding it to Projectile. I think they should be separate things

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
```diff
@@ -1,18 +1,35 @@
...
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import static javax.json.Json.createArrayBuilder;
 import static javax.json.Json.createObjectBuilder;
 
 public final class JsonUtils {
 
+    private static final ObjectMapper mapper = new ObjectMapper();
```
ia3andy: What do you think about naming constant in upper case?

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
```diff
@@ -1,18 +1,35 @@
...
+
     private JsonUtils() {
```
ia3andy: Since we use different Json Libraries, maybe naming it Jacksons or JsonMappers?

ia3andy:
@gastaldi my first guess was to agree with your last comment about separation. However, I find your current solution pretty clean and simple. It seems to me that `Projectile` is not a "data" object but more an object which keeps the runtime state and configuration during the process (it already contains a `Path` for example). 
The only drawback I see about keeping it in the `Projectile` is if it ever get serialized.

edewit:
I must agree that SSE is all we need as we don't need bi directional communication. But websockets are better supported in browsers and it's easier to detect if the browser is still connected. What would be the benefits of changing to SSE?

base/src/main/java/io/fabric8/launcher/base/JsonUtils.java
gastaldi: That's an implementation detail, I think we should leave in the JsonUtils class for now and then separate later if needed

Issue #250, opened 2018-04-03
Scan organization is scanning my github user space even if select github organization


ravsa:
I selected an organization in the launch flow but the Jenkins build scan-organization is scanning my GitHub userspace only.
![1](https://user-images.githubusercontent.com/11839674/38249205-878cc330-3768-11e8-96cc-1dbbc43ffff5.png)

here is the Jenkins [**Scan Organization Log**
](https://gist.github.com/ravsa/e8f6b0ee841e1292ba54a4026d36bba3)
```log

Started by remote host kubernetes with note: Triggered by the change to the ConfigMap
[Tue Apr 03 03:12:42 UTC 2018] Starting organization scan...
[Tue Apr 03 03:12:42 UTC 2018] Updating actions...
Looking up details of ravsa...
Organization URL: Ravindra Singh Ratnawat
[Tue Apr 03 03:12:42 UTC 2018] Consulting GitHub Organization
03:12:42 Connecting to https://api.github.com using ravsa/****** (cd-github)
03:12:43 Looking up repositories of myself ravsa
Ignoring analytics-driven-ui
Ignoring angular2-rxjs-chat
Ignoring api-machine-stacks
Ignoring Automations
Ignoring brainteaser
Ignoring Competitive-Coding
Ignoring data_scraping
Ignoring dec-22-space-12121
Ignoring dec-22-space-123245
.
.
.
[Tue Apr 03 03:12:45 UTC 2018] Finished organization scan. Scan took 2.8 sec
Finished: SUCCESS
```

PR #251, opened 2018-04-03
fix: ensure create jenkins config take github organization name if exist

ravsa:
In new launcher experience jenkins config does not take git organization name only takes github user name and jenkins scan-organization scans github user space only.

fixes: #250


gastaldi:
Good catch @ravsa! Will merge once the CI finishes the build.

PR #264, opened 2018-04-04
fix(osio): /osio/import now clones the repository and performs the necessary changes

### LINKED ISSUE #258: OSIO import should clone the repository and perform necessary changes before completing
The legacy code clones the imported repositories and modifies the pipelines. It doesn't seem we're doing that in `OsioMissionControl#launchImport` 
Not cloning it also causes https://github.com/openshiftio/openshift.io/issues/2358

gastaldi:
Fixes #258


addons/osio-addon/src/main/java/io/fabric8/launcher/osio/OsioMissionControl.java
```diff
@@ -122,6 +118,23 @@ public Boom launch(Projectile genericProjectile) throws IllegalArgumentException
     }
 
 
+    public OsioImportProjectile prepareImport(OsioImportProjectileContext context) {
+        final Space space = witClient.findSpaceById(context.getSpaceId());
+        Path path = gitSteps.clone(context);
+        for (ProjectilePreparer preparer : preparers) {
```
quintesse: We might have to make sure that the preparers are robust enough to handle imports? (For example, what if the imported code doesn't have a README.md? Will the preparer fail, ignore it or create a README?)

gastaldi: They are, the ProjectileContext type is different. In your example, it will ignore

PR #267, opened 2018-04-05
refactor(identity): Remove type from TokenIdentity

### LINKED ISSUE #261: Remove type from TokenIdentity

ia3andy:
We prefer to specify the authorization type directly in the service.

Introduce a notion of standard AuthorizationType to simplify request creation:
    - TokenIdentity -> Bearer
    - UserPasswordIdentity -> Basic

Fixes #261


base/src/main/java/io/fabric8/launcher/base/identity/Identity.java
```diff
@@ -14,14 +14,30 @@
...
+        return createAuthorization(this, getStandardAuthorizationType());
     }
 
+    /**
+     * Get the request authorization for this identity using the specified authorization type
+     *
+     * @param type the {@link AuthorizationType}
+     * @return the request authorization
+     */
+    default String toRequestAuthorization(AuthorizationType type) {
```
gastaldi: Add a @Derived or it will be created as a property afaik

base/src/main/java/io/fabric8/launcher/base/identity/Identity.java
```diff
@@ -14,14 +14,30 @@
...
-    @Value.Lazy
-    default String toRequestHeaderKey() {
-        return createRequestAuthorizationHeaderKey(this);
-    }
+    /**
+     * Get the standard authorization type to use with this identity
+     *
+     * @return the {@link AuthorizationType}
+     */
+    AuthorizationType getStandardAuthorizationType();
```
gastaldi: Why not just getAuthorization?

ia3andy: if I change to `getDefaultAuthorizationType` and `toRequestAuthorization()` do you prefer?

gastaldi: My point is to why add the `Default` or `Standard` in this method name.

ia3andy: because it's just one of the available authorization type to use as standard/default, meaning when you make an http request with a token, bearer is the most common usage and when you user password, basic is.

gastaldi: Ah so better use `Default` instead of `Standard` and add a @Derived or the generated Immutable will be wrong I guess

ia3andy: Done ;)

base/src/main/java/io/fabric8/launcher/base/identity/Identity.java
```diff
@@ -14,14 +14,31 @@
...
-    @Value.Lazy
-    default String toRequestHeaderKey() {
-        return createRequestAuthorizationHeaderKey(this);
-    }
+    /**
+     * Get the default authorization type to use with this identity
+     *
+     * @return the {@link AuthorizationType}
+     */
+    AuthorizationType getDefaultAuthorizationType();
```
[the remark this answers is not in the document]
ia3andy: It's overrided as default in TokenIdentity and in UserPasswordIdentity, I've checked it's not generating any field.

PR #277, opened 2018-04-08
refactor(osio): Merge OSIO clients api and impl packages

gastaldi:
Why?
----
- Current abstraction is broken. Impl is visible for consumers of API 
- It is highly unlikely that the Launcher will need to provide different implementations for these OSIO clients
- API can be defined as public methods and implementation as private methods
- Easer maintenance (no need to maintain the code in a separate interface)

What if I want to mock these services?
----------------------------------------
Use Mockito, which is already included in the base-test/ module


gastaldi:
Hold until I fix 

```
(main) Caused by: org.jboss.weld.exceptions.UnproxyableResolutionException: WELD-001436: Type io.fabric8.launcher.osio.client.OsioWitClient is not proxyable because it has a private constructor private io.fabric8.launcher.osio.client.OsioWitClient() - Managed Bean [class io.fabric8.launcher.osio.client.OsioWitClient] with qualifiers [@Any @Default].
```

PR #278, opened 2018-04-09
feat(core): Documentation readme fetching is part of the runtime

### LINKED ISSUE #252: Make documentation readme fetching part of the runtime
It is currently done during the build using the `
addons/launcher-addon/add-template-files.sh` script.

The process should be close to the one used in the BCS. 

As part of issue #185, it will improve `mvn clean install` perf

ia3andy:
feat(core): Documentation readme fetching is part of the runtime

Add `BoosterDocumentationStore` and test in charge of maintaining (fetch/reload) the documentation files and providing them.

Use `BoosterDocumentationStore` in `BoosterReadmeProcessorImpl`.

Add new endpoint:
    /booster-documentation/reload: to call `BoosterDocumentationStore.reload`.

Add new env 'LAUNCHER_BACKEND_DOCUMENTATION_RELOAD_TOKEN' to set the check token for doc reload.

Add tests.

Fixes #252


core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterDocumentationStoreImpl.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }, async);
+    }
+
+    private static Path cloneGitRepository(final String readmeRepositoryURI, final String branch) throws IOException {
+        logger.log(Level.INFO, "Indexing contents from {0} using {1} ref",
+                   new Object[]{readmeRepositoryURI, branch});
+        final Path catalogPath = Files.createTempDirectory("booster-readme");
+        logger.info("Created " + catalogPath);
+        ProcessBuilder builder = new ProcessBuilder()
```
gastaldi: Any reason to not use Git sparse-checkout here? https://briancoyner.github.io/2013/06/05/git-sparse-checkout.html

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
```diff
@@ -42,6 +43,9 @@
     @Inject
     private BoosterCatalogFactory boosterCatalogFactory;
 
+    @Inject
+    private BoosterDocumentationStore boosterDocumentationStore;
```
gastaldi: I think this should belong in a separate endpoint

pom.xml
```diff
@@ -381,6 +387,12 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
```
gastaldi: mockito is part of base-test, if adding to parent project, maybe it needs to be removed from there?

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterReadmeProcessorImpl.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("all")
+    public Map<String, String> getRuntimeProperties(final String deploymentType, final Mission mission, final Runtime runtime) throws IOException {
+        Properties props = new Properties();
+
+        final Path path = getReadmePropertiesPath(getDocumentationBasePath(), deploymentType, mission.getId(), runtime.getId());
+
+        try (InputStream is = Files.newInputStream(path)) {
```
gastaldi: Check if the path exists and return an empty Map if not?

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterReadmeProcessorImpl.java
```diff
@@ -0,0 +1,74 @@
...
+            props.load(is);
+        }
+
+        Map<String, String> map = (Map) props;
```
gastaldi: Add the SuppressWarning annotation in the code (see the Effective Java, page 124 - Eliminate Unchecked Warnings)

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterDocumentationStoreImpl.java
ia3andy: We talked about it with @quintesse, and though it could wait:
- it was easier to code this way in a first shot (1 command instead of multiple)
- it may change soon depending on the documentation future location

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
ia3andy: Look at the issue, I followed @quintesse idea there

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterReadmeProcessorImpl.java
ia3andy: Yup it is part of the rebase with master I was out of date

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterReadmeProcessorImpl.java
ia3andy: Yep!

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
gastaldi: I think what he meant was to do something similar, It doesn't make sense to trigger the readme index in this endpoint

pom.xml
ia3andy: this is weird, it was not finding it.. ok this is now fixes ;)

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterReadmeProcessorImpl.java
gastaldi: Don't forget to log.warn that

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterDocumentationStoreImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final String DOCUMENTATION_BRANCH = "master";
+
+    private volatile CompletableFuture<Path> pathCompletableFuture;
+
+    private final ExecutorService async;
+
+    private DocumentationFilesDownloader downloader;
+
+    @Inject
+    public BoosterDocumentationStoreImpl(final ManagedExecutorService managedExecutorService) {
```
gastaldi: Use the `java.util.concurrent.ExecutorService` type instead

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterReadmeProcessorImpl.java
```diff
@@ -0,0 +1,74 @@
...
+    public Map<String, String> getRuntimeProperties(final String deploymentType, final Mission mission, final Runtime runtime) throws IOException {
+        Properties props = new Properties();
+
+        final Path path = getReadmePropertiesPath(getDocumentationBasePath(), deploymentType, mission.getId(), runtime.getId());
+
+        try (InputStream is = Files.newInputStream(path)) {
+            props.load(is);
+        }
+
+        @SuppressWarnings("all")
```
gastaldi: This should be `@SuppressWarnings("unchecked")`

web/src/main/java/io/fabric8/launcher/web/endpoints/BoosterCatalogEndpoint.java
quintesse: No problem here with making it a separate endpoint.

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterDocumentationStoreImpl.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private static final String DOCUMENTATION_REPOSITORY = "https://github.com/fabric8-launcher/launcher-documentation.git";
+
+    private static final String DOCUMENTATION_BRANCH = "master";
+
+    private volatile CompletableFuture<Path> pathCompletableFuture;
+
+    private final ExecutorService async;
+
+    private DocumentationFilesDownloader downloader;
```
quintesse: Where does this class/interface come from? I don't see it defined anywhere?

quintesse: Ok, found it, was so tiny I looked over it :)

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/documentation/BoosterDocumentationStoreImpl.java
```diff
@@ -0,0 +1,127 @@
...
+            int exitCode = builder.start().waitFor();
+            assert exitCode == 0 : "Process returned exit code: " + exitCode;
+        } catch (InterruptedException e) {
+            logger.log(Level.WARNING, "Interrupted indexing process");
+            throw new IOException("Interrupted", e);
+        }
+        return catalogPath;
+    }
+
+    public interface DocumentationFilesDownloader {
```
gastaldi: Do you think this interface is needed since it's only used in testing or can it be replaced by a `Supplier<Path>`?

ia3andy: I was thinking of using this one in a later commit to use a local directory directly (if env is set), so I do prefer the second solution ;)

ia3andy:
I splited the test in:
- an IT (to test that the files are there after the clone).
- a unit test set by wrapping the path supplier with the files in test/resources

ia3andy:
@gastaldi can merge?

core/core-api/src/main/java/io/fabric8/launcher/core/api/readme/ReadmeProcessor.java
```diff
@@ -3,6 +3,8 @@
 import java.io.IOException;
 import java.util.Map;
 
+import javax.annotation.Nullable;
```
gastaldi: Can this interface be moved to the documentation package?

ia3andy: I though it was already there.. lol.. fixed

Issue #285, opened 2018-04-10
repository endpoint only returns non forks


gastaldi:
_From @edewit on April 10, 2018 14:52_

The repository endpoint should also include forks
    org.kohsuke.github.GHRepositorySearchBuilder#forks

_Copied from original issue: fabric8-launcher/launcher-planning#154_

PR #286, opened 2018-04-10
fix(git): repository endpoint returns forks as well

ia3andy:
Fixes: #285


alien-ike:
### Ike Plugins (test-keeper)

Thank you @ia3andy for this contribution!



If you are an admin and you are sure that no test is needed then you can use the command `/ok-without-tests` as a comment to make the status green.

PR #309, opened 2018-04-12
feat(health-check): Use Microprofile Health Check

gastaldi:
This replaces the trivial HealthCheckEndpoint

Health checks are used to probe the state of a computing node from another machine (i.e. kubernetes service controller) with the primary target being cloud infrastructure environments where automated processes maintain the state of computing nodes.
In this scenario, health checks are used to determine if a computing node needs to be discarded (terminated, shutdown) and eventually replaced by another (healthy) instance.
It’s not intended (although could be used) as a monitoring solution for human operators.

This allows implementations to return information that could decide whether this service is available or not.

Spec: https://github.com/eclipse/microprofile-health/
Article: https://developers.redhat.com/blog/2018/03/05/cloud-native-microprofile-config-healthcheck-openshift/


alien-ike:
### Ike Plugins (test-keeper)

alien-ike:
Hey @gastaldi! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer, but not pull request creator for this command to take an effect.

gastaldi:
Note that the build ran faster with this PR, probably because I removed HealthCheckEndpointIT

ia3andy:
Good job removing code without loosing functionalities :)

PR #311, opened 2018-04-12
docs(community): Improve PR template

ia3andy:
The title line was not used (the commit msg is used instead).

I changed the first line to point to the whole contributing guide.
I added the `create issue` and `mvn clean install` check.

I removed the space in the checkboxes because [ x] is not working, only [x] is..


ia3andy:
I kept the "I am happy with my contribution" but we can discuss it, I personally like it because it means:
- you are proud of your contribution
- you didn't do something wrong (not justified) in there
- you are happy to have been working with us

PR #314, opened 2018-04-13
chore(template): Try alternative syntax

quintesse:
Using the `${{ }}` syntax with double braces messes with my scripts, let's see
if we can do without.

### DIFF
--- a/openshift/template.yaml
+++ b/openshift/template.yaml
@@ -18,41 +18,34 @@ parameters:
 - name: IMAGE
   value: fabric8/launcher-backend
   required: true
-
 - name: IMAGE_TAG
   value: latest
   required: true
-
 - description: CPU request
   displayName: CPU request
   required: true
   name: CPU_REQUEST
   value: "10m"
-
 - description: CPU limit
   displayName: CPU limit
   required: true
   name: CPU_LIMIT
   value: "1000m"
-
 - description: Memory request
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

bartoszmajsak:
You can create your own `test-keeper.yaml` file to exclude such files  - http://arquillian.org/ike-prow-plugins/#test-keeper-config

```yaml
skip_validation_for:
  - 'openshift/**'
```

put it in the root of the repo.

quintesse:
Would have been nice if that file had been named `.test-keeper.yaml`. We don't need _more_ files in the root of the repo that are only plumbing IMO.

bartoszmajsak:
> Would have been nice if that file had been named .test-keeper.yaml. We don't need more files in the root of the repo that are only plumbing IMO.

You will still have this file there anyway, to start with. But I understand a need for having it separated - how about having `.ike-prow/` folder instead? We are thinking about this location as a home for the config of all the plugins we are cooking.

gastaldi:
> how about having .ike-prow/ folder instead? 
I'm fine with that @bartoszmajsak

bartoszmajsak:
We will work on that and fire PR with cleanup when rolled out. Thanks for being early adopters!

PR #332, opened 2018-04-19
feat(gzip): Enable GZIP compression

### LINKED ISSUE #331: Enable GZIP compression if possible
It would be nice to GZIP the response to save bandwidth when requesting data from the backend

gastaldi:
Why GZip Compress Content?
--------------------------
GZip compressing makes the data sent to the browser smaller. This speeds up the download. This is especially beneficial for mobile phones where internet bandwidth may be limited. GZip compressing content adds a CPU overhead on the server and browser, but it is still speeding up the total page load compared to not GZip compressing.

GZip HTTP Headers
-----------------
The browser includes the `Accept-Encoding` HTTP header in requests sent to an HTTP server (e.g. a Java web server). The content of the `Accept-Encoding` header tells what content encodings the browser can accept. If that header contains the value gzip in it, the browser can accept GZip compressed content. The server can then GZip compress the content sent back to the browser.

If the content sent back from the server is GZip compressed, the server includes the `Content-Encoding` HTTP header with the value gzip in the HTTP response. That way the browser knows that the content is GZip compressed.

Why a GZip Servlet Filter?
--------------------------
It is executed before anything else, so it can be used for any content exposed in this application

Measured Improvements
---------------------

- /api/booster-catalog/missions reduced from 3266 to 1202 bytes
- /api/services/jenkins/pipelines reduced from 9757 to 701 bytes

Fixes #331

### DIFF
--- a/web/src/main/java/io/fabric8/launcher/web/providers/gzip/GZipFilter.java
+++ b/web/src/main/java/io/fabric8/launcher/web/providers/gzip/GZipFilter.java
@@ -1,0 +1,60 @@
+package io.fabric8.launcher.web.providers.gzip;
+
+import java.io.IOException;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.annotation.WebFilter;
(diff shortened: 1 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

gastaldi:
Believe me, I tried both solutions before coming up with this. None of them
worked.

>
>

ia3andy:
LGTM then :)

PR #334, opened 2018-04-19
perf(auth): Support parallel executions of token checks

### LINKED ISSUE #327: Support parallel executions of token checks
### Description

`service/openshift/clusters`in OpenShiftEndpoint should support parallel execution of the IdentityProvider (which cannot be simply put in another thread because they are @RequestScoped)
...

### What should happen?
Faster response time to that endpoint
...

### What happens instead?
Slow 
...

### Do you have any idea why?
Yes, I explained in the description :)
...

ia3andy:
See #327 for more info.

Closes #327


ia3andy:
@quintesse I prefer having the // there for two reasons:
- To avoid having problem with @RequestScoped injected instance which are not accessible from // threads.
- Because maybe sometime or in another implementation we may have a way to request them at once and just remove the need to do // calls which would then be transparent to the caller.

WDYT

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/identity/OsioIdentityProvider.java
```diff
@@ -19,28 +23,33 @@
...
 
     @Override
     public Optional<Identity> getIdentity(final String service) {
+        return getIdentity(authorization, service);
+    }
+
+    @Override
+    public Map<String, Optional<Identity>> getIdentities(final Set<String> services) {
+        final TokenIdentity immutableAuthorization = ImmutableTokenIdentity.copyOf(authorization);
+        return services.stream().parallel()
```
quintesse: Using `parallel` in containers is a potential issue. If you can remember when using the `Future`s, you were obliged to use the app server's executor service, this is because you can't  let Java use it default thread pools. But here, using `parallel()` you're doing that under the hood. So, not good.

So either you would pass an `ExecutorService` to this method, or like I suggested in my other comment, you remove the parallelism here and move it up the call chain far enough until you reach a point where you can easily retrieve the app server's executor service.

gastaldi: You can `@Inject ExecutorService` in the constructor in this case

quintesse:
Well first like I explained in my other comment, you can't use parallelism using default thread pools when running inside containers, bad things happen.

And second is that I think it's just bad design to have a service/provider do a single thing, synchronously and then just add an extra method that does the same multiple times asynchronously and in parallel. It just sounds like: let's anything I need to that interface.

For example there exists an API that does a single HTTP request and waits until the response returns. The same API does not have a method for doing multiple requests in parallel, that's the caller's job.

quintesse:
> To avoid having problem with @requestscoped injected instance which are not accessible from // threads.

Well, pass that manually to it's constructor then. Not everything needs to be injected ;)
(Remember the problem "when you have a hammer everything looks like a nail" hehe)

quintesse:
Ok, changed my mind for the following reasons:

 - The Provider isn't really a service interface (it's not in the `service` module), so isn't a generic API
 - This means the Provider is _very_ specific to the application and how it's built
 - So trying to handle the parallelism at that (very high) level of the application is okay

We _do_ need to handle the problem with the `ExecutorService`, but like @gastaldi mentioned, this can be injected. You'd just have to change from a parallel stream to using `CoimpletableFuture`

quintesse:
This is a very useful function: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...-

core/core-api/src/main/java/io/fabric8/launcher/core/spi/IdentityProviders.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public static Optional<Identity> useDefaultOrElseProvide(final Optional<Identity> defaultIdentity, final IdentityProvider identityProvider, final TokenIdentity authorization, final String provider) {
+        if (defaultIdentity.isPresent()) {
+            return defaultIdentity;
+        } else {
+            return identityProvider.getIdentity(authorization, provider);
+        }
+    }
+
+    public static Map<String, Optional<Identity>> getIdentities(final IdentityProvider identityProvider, final TokenIdentity authorization, final Set<String> services) {
```
gastaldi: This code looks really complicated. Are you sure we need this?

ia3andy: Do you see an alternative? or a better way? (look how it used)

core/core-api/src/main/java/io/fabric8/launcher/core/spi/IdentityProviders.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * An utility class for {@link IdentityProviders}.
+ */
+public final class IdentityProviders {
+
+    private IdentityProviders() {
+        throw new IllegalAccessError("utility class");
+    }
+
+    public static Optional<Identity> useDefaultOrElseProvide(final Optional<Identity> defaultIdentity, final IdentityProvider identityProvider, final TokenIdentity authorization, final String provider) {
```
gastaldi: I don't think we need this method. It can be easily solved by using 
`defaultIdentity.orElseGet(()->identityProvider.getIdentity(authorization, provider))`

core/core-api/src/main/java/io/fabric8/launcher/core/spi/IdentityProviders.java
ia3andy: The issue is that you don't want to loose providerName for the identities

core/core-api/src/main/java/io/fabric8/launcher/core/spi/IdentityProviders.java
ia3andy: indeed!! I forgot to remove it I needed it because getIdentity was throwing checked exceptions (but not anymore) ...

base/src/main/java/io/fabric8/launcher/base/http/HttpClient.java
```diff
@@ -47,10 +50,17 @@ public HttpClient() {
...
-        client = createClient(executorService);
+    public HttpClient(final ExecutorService executorService) {
+        this.client = createClient(requireNonNull(executorService, "executorService must not be null"));
     }
 
+    private HttpClient(final OkHttpClient client) {
+        this.client = requireNonNull(client, "client must be specified.");
+    }
+
+    public static HttpClient createForTest() {
```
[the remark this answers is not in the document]
gastaldi: I was thinking of lazily initializing the client (by having a private getClient() in here and using it in the methods implementation instead of the attribute directly)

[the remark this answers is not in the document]
ia3andy: It's safer this way IMO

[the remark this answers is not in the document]
gastaldi: I think it's easier to use the default constructor than to introduce a factory method here, it saves you from deprecating the default constructor :)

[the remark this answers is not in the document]
ia3andy: I think it's really not easier. It requires a lazy client which is just required because of the proxy. Here I set the client to null for the proxy init, making sure it is never used and managed.

The factory method for test is a cheap price to pay and is at more explicit than the constructor that should never be used. More over I am sure than the real constructor never takes null as an ExecutorService.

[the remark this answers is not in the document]
ia3andy: The main reason of this choice is that I didn't want to add code which wouldn't be needed without the proxy (the lazy loading).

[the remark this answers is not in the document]
gastaldi: I'm not really sure I fully agree with this. It feels weird that a class in the `base/` module has a deprecated method because of CDI, something that the base module shouldn't require to function properly.

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/identity/KeycloakIdentityProvider.java
```diff
@@ -0,0 +1,123 @@
...
+        httpClient = HttpClient.createForTest();
+    }
+
+    @Override
+    public CompletableFuture<Optional<Identity>> getIdentityAsync(final TokenIdentity authorization, final String provider) {
+        requireNonNull(authorization, "authorization must be specified.");
+        requireNonNull(provider, "provider must be specified.");
+
+        String url = buildURL(keyCloakURL, realm, provider);
+        return getToken(url, authorization).thenApply(r -> {
```
gastaldi: `thenApply` seems to be doing nothing here?

web/src/main/java/io/fabric8/launcher/web/api/OpenShiftResource.java
```diff
@@ -87,14 +91,15 @@ public JsonArray getSupportedOpenShiftClusters(@Context HttpServletRequest reque
...
                     .forEach(arrayBuilder::add);
         } else {
-            IdentityProvider identityProvider = this.identityProviderInstance.get();
-            clusters.stream()
+            final Set<String> clusterIds = clusters.stream()
                     .filter(b -> !OSIO_CLUSTER_TYPE.equalsIgnoreCase(b.getType()))
                     .map(OpenShiftCluster::getId)
-                    .forEach(clusterId -> identityProvider.getIdentity(clusterId)
-                            .ifPresent(token -> arrayBuilder.add(clusterId)));
+                    .collect(Collectors.toSet());
```
gastaldi: Don't change OpenShiftResource, it will be removed and should work sequentially.

ia3andy:
@gastaldi we discussed the `HttpClient` default constructor with @edewit  and though it would be nicer if we just make `HttpClient` free from CDI and add a producer in core-impl as you can see in this commit:
https://github.com/fabric8-launcher/launcher-backend/pull/334/commits/339b2ffac0e580320fad4c37e37bdd8275dd4a2a

It solves the default constructor problem and let us keep `base` free of CDI concerns.

WDYT?

ia3andy:
@gastaldi @quintesse to solve the complex code issue we add last week, I used spotify CompletablesFutures library (which can be nice to use in other part of the app btw). The code end up pretty simple:
```Java
            final List<CompletableFuture<ClusterVerified>> futures = new ArrayList<>();
            for (OpenShiftCluster cluster : clusterRegistry.getClusters()) {
                futures.add(identityProviderInstance.get().getIdentityAsync(authorizationInstance.get(), cluster.getId())
                                    .thenApply(identity -> new ClusterVerified(cluster, identity.isPresent())));
            }
            clusters = CompletableFutures.allAsList(futures).get();
```

ia3andy:
CI is blocked by: 
https://github.com/openshiftio/openshift.io/issues/3236

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/ApplicationScopedBeansProducer.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.concurrent.ExecutorService;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Produces;
+
+import io.fabric8.launcher.base.http.HttpClient;
+
+@ApplicationScoped
+public final class ApplicationScopedBeansProducer {
```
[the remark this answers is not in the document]
ia3andy: I created this issue https://github.com/fabric8-launcher/launcher-backend/issues/345
I will fix this whatever solution we decide to use in there.

PR #344, opened 2018-04-21
feat(booster-catalog): Support filtering by a script expression

gastaldi:
This introduces a Predicate<RhoarBooster> implementation that can be activated by providing the `LAUNCHER_BOOSTER_CATALOG_FILTER` env param/system property.

The script must evaluate to a boolean using the `booster` variable, that is an instance of RhoarBooster.

Examples:

- `booster.mission.id == 'rest-http'`: will return only HTTP Rest mission boosters
- `booster.runtime.id == 'spring-boot'`: returns only the Spring Boot boosters
- `booster.mission.id == 'rest-http' && booster.runtime.id` == 'spring-boot': returns only HTTP Rest Spring Boot boosters
- `booster.metadata.istio`: returns only boosters that contains the `istio: true` flag in the booster metadata
- `booster.mission.metadata.istio`: returns only boosters that contains the `istio: true` flag in the mission metadata assigned to the booster

Fixes: https://github.com/fabric8-launcher/launcher-planning/issues/153

### DIFF
--- a/core/core-impl/src/main/java/io/fabric8/launcher/core/impl/catalog/RhoarBoosterCatalogFactory.java
+++ b/core/core-impl/src/main/java/io/fabric8/launcher/core/impl/catalog/RhoarBoosterCatalogFactory.java
@@ -12,6 +12,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Predicate;
 import java.util.logging.Logger;
 
 import javax.annotation.PostConstruct;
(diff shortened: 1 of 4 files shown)


quintesse:
Hey @gastaldi , I like this a lot! The only thing I wondered is: why not make the Predicate part of the BCS? Seems like something that is generically useful, right?

The other thing is security. I had actually hoped to be able to expose something like this publicly, perhaps by allowing you to put filters in the URL of the Launcher. But having access to a full script engine might perhaps be a bit too much. You have any ideas on that perhaps?

gastaldi:
@quintesse yes, those are all valid concerns. I thought of moving to BCS in a later step, which shouldn't impact the code since the class is package-level.

About security, Nashorn offers some security flags that we could adopt: https://stackoverflow.com/questions/20793089/secure-nashorn-js-execution

gastaldi:
@quintesse another solution if you are planning to expose it publicly is, instead of exposing a fully-fledged script engine, use a simpler eval library like comons-jexl (https://commons.apache.org/proper/commons-jexl/reference/syntax.html)

rdruss:
Any chance that this feature can be used to deliver osio specific versions of the boosters as well?

gastaldi:
@rdruss osio specific versions of the boosters? what do you mean? This is usually accomplished by having the `runsOn: osio` in the `booster.yaml` metadata.

So, in this case, a possible solution would be to start the backend with a filter expression like: 
`booster.metadata.runsOn == 'osio' || booster.metadata.runsOn.contains('osio')`

rdruss:
@gastaldi, but can you have different versions of the same booster - one that has runsOn set to osio, the others say !osio?

ALRubinger:
@rdruss Yes :)

Issue #354, opened 2018-04-24
Change IdentityProvider.getIdentity behavior


ia3andy:
as @quintesse said:
> I actually think returning Optional here instead of the result or an exception is throwing away useful information. IMO service APIs should never decide for you that errors are not important to you and only log them. A service API should just pass on the error and let the caller decide. And especially when using the CompletableFuture, suddenly you've lost the ability to use the .exceptionally() to distinguish between correct and erroneous results.

New behavior:
- returns the token when the provider is connected
- returns empty when the authorization token is valid but the provider is disconnected 
- throws a checked IdentityProviderException when something else happens

is this logic ok for you guys?

Since this method is called in different part of the app, it will be easier to change/test when the old endpoints will be deleted.

All IdentityProvider impl have to be tested in all those cases.
Every endpoint using this method have to be tested in all those cases too.

gastaldi:
Closing, the IdentityProvider behavior doesn't need to change right now.

PR #356, opened 2018-04-24
fix(osio): make OsioIdentityProvider return empty when an error occurs

ia3andy:
- Make OsioIdentityProvider like KeycloakIdentityProvider by always returning empty in case of error. We should change this when solving #354.
- Make OsioIdentityProvider more strict by only accepting "github" or "openshift-v3" as provider.
- Add urlEncode to Requests and use it in Osio clients.


gastaldi:
Although I approved it, it would be nice to make sure the build in semaphore passes first ;)

Issue #351, opened 2018-04-23
ProcessingException: com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.lang.St...


gastaldi:
https://errortracking.prod-preview.openshift.io/openshift_io/fabric8-launcher-backend/issues/3494/

```
JsonMappingException: Can not deserialize instance of java.lang.String out of START_OBJECT token
 at [Source: org.jboss.resteasy.client.jaxrs.internal.ClientResponse$InputStreamWrapper@30fe9cc; line: 1, column: 2] (through reference chain: java.util.ArrayList[0])
    at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:270)
    at com.fasterxml.jackson.databind.DeserializationContext.reportMappingException(DeserializationContext.java:1234)
    at com.fasterxml.jackson.databind.DeserializationContext.handleUnexpectedToken(DeserializationContext.java:1122)
    at com.fasterxml.jackson.databind.DeserializationContext.handleUnexpectedToken(DeserializationContext.java:1075)
    at com.fasterxml.jackson.databind.deser.std.StdDeserializer._parseString(StdDeserializer.java:862)
...
(89 additional frame(s) were not displayed)

ProcessingException: com.fasterxml.jackson.databind.JsonMappingException: Can not deserialize instance of java.lang.String out of START_OBJECT token
 at [Source: org.jboss.resteasy.client.jaxrs.internal.ClientResponse$InputStreamWrapper@30fe9cc; line: 1, column: 2] (through reference chain: java.util.ArrayList[0])
    at org.jboss.resteasy.client.jaxrs.internal.ClientResponse.readFrom(ClientResponse.java:300)
    at org.jboss.resteasy.client.jaxrs.internal.ClientResponse.readEntity(ClientResponse.java:196)
    at org.jboss.resteasy.specimpl.BuiltResponse.readEntity(BuiltResponse.java:226)
    at io.fabric8.launcher.addon.MissionControl.lambda$getOpenShiftClusters$0(MissionControl.java:185)
    at io.fabric8.launcher.addon.MissionControl.perform(MissionControl.java:216)
...
(76 additional frame(s) were not displayed)

Error while returning openshift clusters
```

This was caused by your recent commit. Mind checking this out?

ia3andy:
[...]

PR #357, opened 2018-04-25
fix: removed duplicate null check in TokenIdentity.fromBearerAuthorizationHeader

gastaldi:
Authorizations.isBearerAuthentication already checks for null values and the test throws a better exception message.

This is	an attempt to figure out what is causing issue #351

### DIFF
--- a/base/src/main/java/io/fabric8/launcher/base/identity/TokenIdentity.java
+++ b/base/src/main/java/io/fabric8/launcher/base/identity/TokenIdentity.java
@@ -24,7 +24,6 @@ public interface TokenIdentity extends Identity {
     }
 
     static TokenIdentity fromBearerAuthorizationHeader(String authorizationHeader) {
-        requireNonNull(authorizationHeader, "authorizationHeader must be specified.");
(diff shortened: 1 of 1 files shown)


base/src/main/java/io/fabric8/launcher/base/identity/TokenIdentity.java
```diff
@@ -24,7 +24,6 @@ default void accept(IdentityVisitor visitor) {
     }
 
     static TokenIdentity fromBearerAuthorizationHeader(String authorizationHeader) {
-        requireNonNull(authorizationHeader, "authorizationHeader must be specified.");
```
ia3andy: it was there to separate NPE when arg is null from IllegalArgumentException when it's incorrect.

ia3andy:
I think you should revert this commit, because at the end it's what let you easily find where the bug is coming from.

ia3andy:
@gastaldi indeed :), I just clicked on the link from the reference in the issue pointing to this PR..

PR #363, opened 2018-04-25
fix(osio): create OpenShift secrets instead of talking to Jenkins

### LINKED ISSUE #362: Do not talk to Jenkins and create credentials in Openshift instead
https://github.com/fabric8-launcher/launcher-backend/blob/master/addons/legacy-osio-addon/src/main/java/io/fabric8/forge/generator/kubernetes/CreateBuildConfigStep.java#L496-L522 needs to be done instead of talking to Jenkins directly, because Jenkins can be idle

Parent issue: https://github.com/openshiftio/openshift.io/issues/3077

gastaldi:
Old launcher does two things in attempt to setup credentials:

- talk directly to jenkins to setup github credentials
- setup secret to get synced by openshift sync plugin

New launcher only does the first thing.

And talking directly to jenkins will fail if jenkins have been idled or for some reason haven't been able to startup fast enough. 
Thus when using new launcher which clickparty and automated tests we end up with jenkins having no secret and jenkins default behavior is to then use anonymous access.
Suggested fix is that new launcher should just do step 2 so sync will always do the work. 

Fixes #362


alien-ike:
### Ike Plugins (test-keeper)

gastaldi:
@aslakknutsen can you review this too?

ldimaggi:
This issue was surfaced in the click party - and also in automated tests - as the users and tests followed a sequence of resetting the user's environment before attempting to create a new space + app with the launcher. It must have been the case than in many instances Jenkins had not started and anonymous access was used.

PR #364, opened 2018-04-25
chore(env): echo all environment vars when sourcing `launcher-env-template.sh`

ia3andy:


ia3andy:
@gastaldi totally agree after the summit ? :p

ia3andy:
https://github.com/fabric8-launcher/launcher-backend/issues/365

PR #372, opened 2018-04-30
fix(mission-control): make sure that the catalog index is finished before proceeding

gastaldi:


alien-ike:
### Ike Plugins (test-keeper)

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/MissionControlImpl.java
```diff
@@ -69,6 +69,9 @@ public CreateProjectile prepare(ProjectileContext context) {
         java.nio.file.Path path;
         try {
             path = Files.createTempDirectory("projectDir");
+            // Wait for index to finish before querying the catalog
+            catalogFactory.waitForIndex();
```
quintesse: I'm not sure I understand why this was necessary? How can anyone ever launch a project before the catalog was indexed? I mean you need to have an index to be able to select the booster you want to launch.

gastaldi: When the backend is redeployed while the user is browsing the frontend, there is a slight chance that the booster isn't indexed yet, this should ensure that the catalog is indexed completely before doing that.

PR #400, opened 2018-05-16
fix(test): Remove hoverfly.jks truststore

ia3andy:
Our HttpClient trust all certificate. We don't need it anymore.


services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceTest.java
```diff
@@ -550,17 +552,21 @@ public void cloneRepository() throws Exception {
 
     private String getRawFileContent(final String fullRepoName, final String fileName) throws IOException {
         URI readmeUri = URI.create(getRawFileUrl(fullRepoName, fileName));
+        HttpClient.create();
```
gastaldi: Why this call?

base/src/main/java/io/fabric8/launcher/base/http/HttpClient.java
```diff
@@ -157,19 +162,42 @@ public void onResponse(Call call, Response response) {
         }
     }
 
-    private static final TrustManager[] trustAllCerts = {
-            new X509TrustManager() {
+    public static final TrustManager[] trustAllCerts = {
```
gastaldi: Not sure I like this. I mean it's a static array being exposed as a public attribute :)

gastaldi:
I merged some interesting parts of this PR. I am not sure about the Hoverfly change though.

I'm closing for now, let's revisit this once we get this cleared out.

PR #412, opened 2018-05-24
feat(core): ActiveMQ Artemis as a new StatusMessageEventBroker impl

gastaldi:


core/core-impl/src/main/java/io/fabric8/launcher/core/impl/events/ArtemisStatusMessageEventBroker.java
```diff
@@ -0,0 +1,98 @@
...
+    private final ActiveMQConnectionFactory activeMQConnectionFactory;
+
+    private final JMSContext jmsContext;
+
+    private final Map<UUID, JMSConsumer> consumers = new ConcurrentHashMap<>();
+
+    public ArtemisStatusMessageEventBroker(String url, String user, String password) {
+        activeMQConnectionFactory = new ActiveMQConnectionFactory(url, user, password);
+        jmsContext = activeMQConnectionFactory.createContext();
+        jmsContext.setClientID(EnvironmentSupport.INSTANCE.getEnvVarOrSysProp("HOST_NAME", "localhost"));
```
quintesse: `HOST_NAME`? Or `HOSTNAME`?

gastaldi: Thanks, I was unsure of the environment var name used in K8s

quintesse: Well, I don't know about K8s, but it's what my local Fedora system uses ^^

PR #417, opened 2018-05-30
Added generic "property filtering" to `/missions` and `/runtimes`

quintesse:


core/core-impl/src/test/java/io/fabric8/launcher/core/impl/catalog/ScriptingRhoarBoosterPredicateTest.java
```diff
@@ -1,5 +1,9 @@
 package io.fabric8.launcher.core.impl.catalog;
```
gastaldi: This class needs to be moved to BCS

gastaldi:
It would be nice to have a test in BoosterCatalogEndpointIT with this new feature as well

alien-ike:
### Ike Plugins (test-keeper)

core/core-impl/src/test/java/io/fabric8/launcher/core/impl/catalog/ScriptingRhoarBoosterPredicateTest.java
quintesse: Done

PR #419, opened 2018-05-31
New `/catalog` endpoint and automatic `X-App` filtering

quintesse:


alien-ike:
### Ike Plugins (test-keeper)

gastaldi:
How about publishing in `/booster-catalog` instead of `/booster-catalog/catalog`?

quintesse:
How do I edit the swagger.yaml file? By hand?? Then how do I check it's correct?

quintesse:
> How about publishing in /booster-catalog instead of /booster-catalog/catalog?

Dunno, because it seems weird to me? We still have other endpoints like `/reindex` there as well.

gastaldi:
IMHO that would be more appropriate (eg. To return the booster catalog you do a `GET /booster-catalog`) 

Any other action (like reindex) is already modeled as subpaths

gastaldi:
I used https://editor.swagger.io/?url=https://forge.api.prod-preview.openshift.io/swagger.yaml to check it's correctness but I think there is an IDE plug-in for that

PR #512, opened 2018-08-24
4216 potential resource leaks osiowitclient

tinakurian:
### Why

Fixing potential resource leaks
Using static final variables
Logging exceptions instead of swallowing them

### Checklist

- [ ] I followed the contribution [guidelines](https://raw.githubusercontent.com/fabric8-launcher/launcher-backend/master/CONTRIBUTING.md).
- [ ] I created an [issue](https://github.com/fabric8-launcher/launcher-backend/issues/new) and used it in the commit message.
- [ ] I have built the project locally prior to submission with `mvn clean install`.
- [ ] I kept a clean commit log (no unnecessary commits, no merge commits).
- [ ] I am happy with my contribution.


alien-ike:
### Ike Plugins (test-keeper)

Thank you @tinakurian for this contribution!



For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/OsioWitClient.java
```diff
@@ -37,16 +37,41 @@
 @RequestScoped
 public class OsioWitClient {
 
+    protected static final String ERROR_HTTP_CLIENT_MUST_BE_SPECIFIED = "httpClient must be specified"; //$NON-NLS-1$
```
gastaldi: `protected` is only used if the class is intended to be extended (which isn't the case here)

tinakurian: Sorry, correct, this can be made package protected. Will update!

edewit:
Making constants out of strings that are only used one time doesn't make sense IMO

bartoszmajsak:
I like the improvement of logging errors! That's a good thing.

I believe howerver that these should be two different PRs - one cleaning up recurring strings and the other fixing a resource leak which is now hard to find. I also doubt if in this case that actually holds true, as `executeAndConsume` does `try-with-resources` on `Response` object which implictly closes `BodyResponse` stream).

Generally speaking, I don't feel like making constants from all the strings we have (which are mostly used once right now) improves readability. Plus we add `//$NON-NLS-1$` which is only to make Eclipse happy.

bartoszmajsak:
@tinakurian you need some changes in the code to make the PR build pass.

tinakurian:
> I like the improvement of logging errors! That's a good thing.

I logged a defect for fixing exception handling so those changes should probably be done there

> I believe however that these should be two different PRs - one cleaning up recurring strings and the other fixing a resource leak which is now hard to find. I also doubt if in this case that actually holds true, as executeAndConsume does try-with-resources on Response object which implictly closes BodyResponse stream. This brought my attention to this class and I think we might have potential leaks on the async consumption of responses. PR coming shortly (#513).

Agreed, i've moved all of the work unrelated to resource leaks out. 

Actually in OsioWitClient, the response gets closed in multiple places. It also gets closed if response.body().string() gets called. Removed the try-with-resources from OsioWitClient as there is no benefit in closing it a third time.

Sounds good. 

> Generally speaking, I don't feel like making constants from all the strings we have (which are mostly used once right now) improves readability. Plus we add //$NON-NLS-1$ which is only to make Eclipse happy.

Many of the constants i made appear like they are only being used in one place - but this is not true. I just didn't want to touch all the other files to use the constants I had created until we do work in those files. Furthermore, creating constants for segments is good practice because it's clear that they shouldn't be touched unless we know exactly what we are doing. Lastly, it's nice to create constants for error messages, since those can b re-used in the future. However, we can follow DRY and make those constants if/when needed rather than doing it upfront. Anyways, those are removed for now... 

RE: NON-NLS - there is apparently a different way to do this in Intellij but since we may not all be using the same editor perhaps this doesn't make sense.

tinakurian:
@edewit Thanks for taking the time to review the PR!

> Making constants out of strings that are only used one time doesn't make sense IMO

This was unclear in the PR earlier. Many of the constants could have actually been used in other places in our source code. Also the hand-generated JSON should probably be created using org.json since we already import this library and it is less error prone than doing it manually. The constants could have been used here as well.

services/openshift-service-impl/src/main/java/io/fabric8/launcher/service/openshift/impl/OpenShiftController.java
```diff
@@ -22,8 +22,7 @@
 
     @Override
     public boolean checkNamespace(String namespaceName) {
-        OpenShiftClient client = getOpenShiftClientOrNull();
-        try {
+        try (OpenShiftClient client = getOpenShiftClientOrNull()) {
```
gastaldi: Closing the openshift client here is not a good idea, as it will be needed for other methods. This class will be removed soon, so don't bother changing it

gastaldi:
@tinakurian can you rebase with master and reformat `JenkinsConfigParserTest` (I think it's missing a space in the `try` block) ?

bartoszmajsak:
@tinakurian once you address all the feedback feel free to remove "do not merge" label. As for rebasing - simply clicking "update branch" here will do.

gastaldi:
@tinakurian it appears that there are conflicts still in this PR. Do you mind rebasing all your commits into a single one in order to resolve them?

gastaldi:
Thanks!

tinakurian:
@gastaldi thanks for merging :-)

PR #557, opened 2018-10-15
Token validation for OSIO auth and Keycloak (standalone launcher)

### LINKED ISSUE #556: JWT singature not validated
### Description

Backend does not validate the signature of the JWT.  This is potentially a big security hole. At this point of time we are calling other services which are validating JWTs, so we are somewhat on the safe side, but we just know that's the case - we should validate every token we receive immediately.

In addition this is requirement from the OSIO Auth team (cc @sbose78)

https://github.com/fabric8-launcher/launcher-backend/blob/14e5c48e62f3c715f08891ea8286e683e9a6c792/core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java#L66-L69

### What should happen?

Token signature should be validated using public keys.

### What happens instead?

We just happily pass the token thru. Because

yolo-cat

### Do you have any idea why?

giphy

### Error log

```
not witnessed ( ͡ ͡° ͡° ʖ ͡° ͡°)
```

bartoszmajsak:
### Why

Backend does not validate the signature of the JWT.  This is potentially a big security hole. At this point of time we are calling other services which are validating JWTs, so we are somewhat on the safe side, but we just know that's the case - we should validate those we receive immediately.

### Description

- introduces JWT validator used in the `SecuredFilter`
- introduces `PublicKeyProvider` SPI with Keycloak and OSIO Auth implementation
- Hoverfly Launcher rule extended to load `shared.json` (class-wide fixture) if no `.json` matching method name found

Closes #556

### Checklist

- [x] I followed the contribution [guidelines](https://raw.githubusercontent.com/fabric8-launcher/launcher-backend/master/CONTRIBUTING.md).
- [x] I created an [issue](https://github.com/fabric8-launcher/launcher-backend/issues/new) and used it in the commit message.
- [x] I have built the project locally prior to submission with `mvn clean install`.
- [x] I kept a clean commit log (no unnecessary commits, no merge commits).
- [x] I am happy with my contribution.

### Integration tests
- [x] OSIO 
- [ ] Standalone Launcher **

** need some help here @gastaldi


addons/osio-addon/pom.xml
```diff
@@ -32,6 +32,13 @@
       <artifactId>launcher-core-impl</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.fabric8.launcher</groupId>
+      <artifactId>launcher-core-impl</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
```
bartoszmajsak: Sharing `TokenFixtures` for tests across the modules.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
```diff
@@ -0,0 +1,90 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+
+import static io.fabric8.kubernetes.client.utils.URLUtils.pathJoin;
+import static io.fabric8.launcher.base.http.Requests.securedRequest;
+import static io.fabric8.launcher.osio.OsioConfigs.getAuthUrl;
+import static java.util.Objects.requireNonNull;
+
+@RequestScoped
```
bartoszmajsak: B/c I'm injecting tokens for each req.

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java
```diff
@@ -55,16 +59,16 @@ public void filter(ContainerRequestContext requestContext) {
...
-
-        // Abort the filter chain with a 401 status code response
-        requestContext.abortWith(
-                Response.status(Response.Status.UNAUTHORIZED)
-                        .build());
+    private DecodedJWT validateToken(String token) {
+        final DecodedJWT jwt = JWT.decode(token);
+        final JWTValidator jwtValidator = new JWTValidator(jwt.getIssuer(), publicKeyProvider);
+        if (!jwtValidator.validate(token)) {
+            throw new IllegalArgumentException("Invalid token");
```
bartoszmajsak: @gastaldi Shall we pass it as error to Sentry? :fire: :fire: :bomb: :fireworks:

core/core-impl/src/test/java/io/fabric8/launcher/core/impl/filters/TokenFixtures.java
```diff
@@ -0,0 +1,53 @@
+package io.fabric8.launcher.core.impl.filters;
+
+public class TokenFixtures {
```
gastaldi: If this class is meant to be used in the tests only, move it to launcher-base-test

web/src/main/java/io/fabric8/launcher/web/producers/PublicKeyProviderProducer.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+import io.fabric8.launcher.core.spi.Application;
+import io.fabric8.launcher.core.spi.PublicKeyProvider;
+
+import static io.fabric8.launcher.core.spi.Application.ApplicationLiteral.of;
+import static io.fabric8.launcher.core.spi.Application.ApplicationType.valueOf;
+
+/**
+ * @author <a href="mailto:ggastald@redhat.com">George Gastaldi</a>
```
gastaldi: You created dis? I created dis!

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java
gastaldi: Doesn't look like it needs to. If the token is invalid, there isn't much we can do about it

core/core-impl/src/test/java/io/fabric8/launcher/core/impl/filters/TokenFixtures.java
bartoszmajsak: Fixed with 1b3cf1de

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
gastaldi: I don't see any state in this class, I believe it's safe to be `ApplicationScoped`

bartoszmajsak:
@gastaldi does the semaphore build run tests against KC? we need to cover this path of validation. OSIO seems to work fine

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
gastaldi: Ah ignore my comment, TokenIdentity is there

gastaldi:
> @gastaldi does the semaphore build run tests against KC? we need to cover this path of validation. OSIO seems to work fine

@bartoszmajsak no, it does not. AFAIK the existing KC tests are run against Hoverfly, but I may be wrong

bartoszmajsak:
@gastaldi right, that's what I have provided. Can you verify that branch against Launcher Standalone?

gastaldi:
@bartoszmajsak there is a 3rd scenario which is Launcher without Keycloak (maybe that's why we haven't implemented token validation yet).

Right now when that happens, the frontend sends a dummy Authorization token (readable in JWT Format)

bartoszmajsak:
> @bartoszmajsak there is a 3rd scenario which is Launcher without Keycloak (maybe that's why we haven't implemented token validation yet).
> 
> Right now when that happens, the frontend sends a dummy Authorization token (readable in JWT Format)

How this should be handled though and when is this scenario valid?

gastaldi:
>How this should be handled though and when is this scenario valid?

When the KeycloakService resolution is unsatisfied, see https://github.com/fabric8-launcher/launcher-backend/blob/master/web/src/main/java/io/fabric8/launcher/web/producers/IdentityProviderProducer.java

bartoszmajsak:
In https://github.com/fabric8-launcher/launcher-backend/pull/557/commits/b7e4f79819c138e0f5716705b3d280b30a57d1ef I introduced `HttpClient.getContent` static method to handle body extraction (used in both public key providers). WDYT?

core/core-impl/src/test/java/io/fabric8/launcher/core/impl/identity/KeycloakPublicKeyProviderHoverflyTest.java
```diff
@@ -0,0 +1,95 @@
...
+                               .andForSimulationOnly(LAUNCHER_KEYCLOAK_REALM.propertyKey(), "rh-developers-launch"))
+            .around(HOVERFLY_RULE);
+
+    @Rule
+    public LauncherPerTestHoverflyRule hoverflyPerTestRule = new LauncherPerTestHoverflyRule(HOVERFLY_RULE);
+
+    @Test
+    public void should_receive_key_based_on_its_kid() {
+        // given
+        final PublicKeyProvider publicKeyProvider = new KeycloakPublicKeyProvider(ImmutableKeycloakParameters.builder().build(), getKeycloakToken(), HttpClient.create());
```
tinakurian: Perhaps this can go in a beforeEach? Possibly even a beforeClass instead?

bartoszmajsak:
Tasks based on the video chat with @gastaldi 

- [ ] adjust call to keycloak 3.2 (using `GET /realms/{realm}/protocol/openid-connect/certs`)
- [ ] test changes against standalone launcher
- [ ] pass through dummy token when keycloak is not configured (for local dev or on-prem without auth) - based on @gastaldi review

base/pom.xml
```diff
@@ -20,6 +20,11 @@
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>okhttp</artifactId>
     </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>${version.commons.codec}</version>
```
gastaldi: Move it to the parent pom

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java
```diff
@@ -42,29 +56,40 @@ public void filter(ContainerRequestContext requestContext) {
...
+        requestContext.setProperty("USER_NAME", securityContext.getUserPrincipal().getName());
+        requestContext.setSecurityContext(securityContext);
+    }
 
-        // Abort the filter chain with a 401 status code response
-        requestContext.abortWith(
-                Response.status(Response.Status.UNAUTHORIZED)
-                        .build());
+    // We do not validate tokens in case no keycloak linked for standalone launcher
+    private boolean shouldValidate(HttpServletRequest request) {
```
bartoszmajsak: @gastaldi here's how I decide if we should validate or not

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
```diff
@@ -23,8 +25,19 @@
 public @interface Application {
 
     enum ApplicationType {
+
         LAUNCHER,
-        OSIO
+        OSIO;
+
+        public static ApplicationType fromHeader(HttpServletRequest request) {
```
bartoszmajsak: @gastaldi moved that method to the type itself as it's used in both app-aware producer and token validation

core/core-impl/pom.xml
```diff
@@ -54,4 +54,47 @@
...
+            </configuration>
+          </execution>
+          <execution>
+            <id>no-token-validation</id>
+            <goals>
+              <goal>integration-test</goal>
+            </goals>
+            <configuration>
+              <skip>false</skip>
+              <test>NoKeycloakUrlSetSecuredIT</test>
```
bartoszmajsak: @gastaldi this is how I run test where bypassing token validation is expected. WDYT?

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java
gastaldi: LGTM

core/core-impl/pom.xml
gastaldi: That looks fine. How about setting the environment var inside the test? Each test should already run in its own process afaik

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
gastaldi: The HttpServletRequest in there kinda triggers me the core with a strong dependency on the Servlet API.

gastaldi: But maybe that's ok, WDYT @quintesse ?

bartoszmajsak:
Tests failing due to

```
ost http://geminiservice.analytics/api/v1/user-repo/drop: dial tcp: lookup geminiservice.analytics on 172.28.57.27:53: no such host\nPost http://geminiservice.analytics/api/v1/user-repo/drop: dial tcp: lookup geminiservice.analytics on 172.28.57.27:53: no such host\n","status":"500","title":"Internal error"}]}
```

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
bartoszmajsak: Yeah this is a bit of a smell. On the other hand it's where we use that thing anyway. Alternatively one can make yet another `Util` living somewhere closer to servlet world.

base/pom.xml
bartoszmajsak: I just removed the version - that should do. 41d33a11

core/core-impl/pom.xml
bartoszmajsak: In theory, we can have a test which programmatically invokes this test class... but I am not sure that's a better solution, to be honest. What do we decide? Is that path good enough?

bartoszmajsak:
@gastaldi I think this is now ready for the final review

base/src/main/java/io/fabric8/launcher/base/http/HttpClient.java
```diff
@@ -48,6 +48,28 @@ public static HttpClient create() {
...
+     * @return "null" so that it can be understood by Json parser as null node
+     * @throws IOException
+     */
+    public static String getContent(ResponseBody body) throws IOException {
+        if (body == null) {
+            return "null";
+        }
+        final String content;
+        final String bodyAsString = body.string();
+        if (bodyAsString == null || bodyAsString.trim().isEmpty()) {
```
tinakurian: minor-- if we assign content like so: final String content = "nul" then we can negate the if statement,  assign content to bodyAsString in the if clause and eliminate the else clause

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
```diff
@@ -0,0 +1,57 @@
...
+            key = stripHeaderAndFooter(key);
+            final X509EncodedKeySpec keySpecX509 = new X509EncodedKeySpec(java.util.Base64.getDecoder().decode(key));
+            return (RSAPublicKey) kf.generatePublic(keySpecX509);
+        } catch (InvalidKeySpecException e) {
+            throw new IllegalArgumentException("Unable to create RSA Public Key", e);
+        }
+    }
+
+    private static String stripHeaderAndFooter(String key) {
+        key = key.replaceAll("\\n", "")
```
tinakurian: instead of assigning to a param variable it is better just to return key.replaceAll(...)...

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java
```diff
@@ -42,29 +56,40 @@ public void filter(ContainerRequestContext requestContext) {
...
-            // Set the user name as a request property
-            requestContext.setProperty("USER_NAME", securityContext.getUserPrincipal().getName());
-            requestContext.setSecurityContext(securityContext);
+            final DecodedJWT jwt = JWT.decode(token);
+            if (shouldValidate(request)) {
+                validateToken(jwt);
+            }
+            propagateSecurityContext(requestContext, jwt);
 
         } catch (Exception e) {
```
tinakurian: don't we lose the stack trace here since nothing is being done with e?

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/identity/KeycloakPublicKeyProvider.java
```diff
@@ -0,0 +1,110 @@
...
+                                      final RSAPublicKey publicKey = transformToRsa(keyNode);
+                                      publicKeys.put(extractFieldFromNodeOrDefaultTo(keyNode, "kid", "kid"), publicKey);
+                                  }
+                );
+        return publicKeys;
+    }
+
+    private static RSAPublicKey transformToRsa(JsonNode keyNode) {
+        final String modulus = extractFieldFromNodeOrDefaultTo(keyNode, "n", null);
+        final String exponent = extractFieldFromNodeOrDefaultTo(keyNode, "e", null);
```
tinakurian: these two values are only used if we didn't fail with the exception - maybe we should move it down to the place where they are being used so we don't calculate them unnecessarily upon failure

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/identity/KeycloakPublicKeyProvider.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+            return findAllPublicKeys(node);
+        } catch (final IOException e) {
+            throw new IllegalStateException("Error while fetching token from keycloak", e);
+        }
+    }
+
+    private static Map<String, RSAPublicKey> findAllPublicKeys(JsonNode node) {
+        final Map<String, RSAPublicKey> publicKeys = new HashMap<>();
+        if (!node.hasNonNull("keys")) {
```
tinakurian: publicKeys is only used if we didn't return from the if clause - maybe we can move it below so we aren't unnecessarily allocating a hashmap

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
```diff
@@ -0,0 +1,94 @@
...
+
+import static io.fabric8.kubernetes.client.utils.URLUtils.pathJoin;
+import static io.fabric8.launcher.base.http.HttpClient.getContent;
+import static io.fabric8.launcher.base.http.Requests.securedRequest;
+import static io.fabric8.launcher.osio.OsioConfigs.getAuthUrl;
+import static java.util.Objects.requireNonNull;
+
+@RequestScoped
+@Application(Application.ApplicationType.OSIO)
+public class AuthPublicKeyProvider implements PublicKeyProvider {
```
tinakurian: it may be good if we documented public classes

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
```diff
@@ -0,0 +1,57 @@
...
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private RSAPublicKeyConverter() {
+        throw new IllegalAccessError("Utility class");
+    }
+
+    public static RSAPublicKey fromJWT(String mod, String exp) {
```
tinakurian: may be beneficial to document public api

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            final BigInteger modulus = new BigInteger(1, Base64.decodeBase64(mod));
+            final BigInteger exponent = new BigInteger(1, Base64.decodeBase64(exp));
+            return (RSAPublicKey) kf.generatePublic(new RSAPublicKeySpec(modulus, exponent));
+        } catch (InvalidKeySpecException ex) {
+            throw new IllegalStateException("Failed converting JWK to RSA", ex);
+        }
+    }
+
+    public static RSAPublicKey fromString(String key) {
```
tinakurian: may be beneficial to document public api

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
tinakurian: may be beneficial to document public api

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
bartoszmajsak: What would you like to have mention in the doc what is not quickly understandable from the code itself? Looking for a hint.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
```diff
@@ -0,0 +1,94 @@
...
+                throw new IllegalStateException(extractFieldFromNodeOrDefaultTo(node, "errors", ""));
+            }
+            return findAllPublicKeys(node);
+        } catch (final IOException e) {
+            throw new IllegalStateException("Error while fetching token from OSIO auth service", e);
+        }
+    }
+
+    private static Map<String, RSAPublicKey> findAllPublicKeys(JsonNode node) {
+        final Map<String, RSAPublicKey> publicKeys = new HashMap<>();
```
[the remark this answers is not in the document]
bartoszmajsak: Sure

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
bartoszmajsak: Fixed in 44f01288, thanks

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
bartoszmajsak: What kind of information would like to see here?

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/client/AuthPublicKeyProvider.java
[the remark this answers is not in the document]
bartoszmajsak: Fixed in 7879ed0b, thanks

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/identity/KeycloakPublicKeyProvider.java
bartoszmajsak: Fixed in a1b61836. Thanks.

base/src/main/java/io/fabric8/launcher/base/http/HttpClient.java
bartoszmajsak: That potentially makes it more concise. The benefit of the current approach is that having `String content` as `final` guarantees that you can't change it later on, only in `if-else` block. Having said that I reworked this method based on your suggestion. Shipped in 8b79d053

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.math.BigInteger;
+import java.security.KeyFactory;
+import java.security.NoSuchAlgorithmException;
+import java.security.interfaces.RSAPublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.RSAPublicKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+
+import org.apache.commons.codec.binary.Base64;
```
gastaldi: How about using [`java.util.Base64`](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html) instead of adding a dependency to commons-codec?

core/core-impl/pom.xml
gastaldi: If that works, I'm fine with this solution :+1:

core/core-impl/src/main/java/io/fabric8/launcher/core/impl/filters/SecuredFilter.java
```diff
@@ -42,29 +61,40 @@ public void filter(ContainerRequestContext requestContext) {
...
-    private void abortWithUnauthorized(ContainerRequestContext requestContext) {
+    private void propagateSecurityContext(ContainerRequestContext requestContext, DecodedJWT jwt) {
+        final JWTSecurityContext securityContext = new JWTSecurityContext(jwt);
+        requestContext.setProperty("USER_NAME", securityContext.getUserPrincipal().getName());
+        requestContext.setSecurityContext(securityContext);
+    }
+
+    // We do not validate tokens in case no keycloak linked for standalone launcher
+    private boolean shouldValidate(HttpServletRequest request) {
+        System.out.println(LAUNCHER_KEYCLOAK_URL.value());
```
gastaldi: Remove this line

bartoszmajsak:
```suggest
```

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static final String PUBLIC_KEY_ALGORITHM = "RSA";
+
+    private static final KeyFactory kf;
+
+    static {
+        try {
+            kf = KeyFactory.getInstance(PUBLIC_KEY_ALGORITHM);
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
tinakurian: may be good to define and throw a dedicated exception instead of using runtimeexception, exception, error, etc..

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
```diff
@@ -23,8 +25,19 @@
...
+            // If X-App is not specified, assume fabric8-launcher
+            final String app = Objects.toString(request.getHeader("X-App"), "launcher").toUpperCase();
+            try {
+                return valueOf(app);
+            } catch (IllegalArgumentException iae) {
+                throw new IllegalArgumentException("Header 'X-App' has an invalid value: " + app);
```
tinakurian: should we maybe use the value iae so we don't swallow it?

base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherPerTestHoverflyRule.java
```diff
@@ -16,24 +19,38 @@ public LauncherPerTestHoverflyRule(HoverflyRule hoverflyRule) {
...
-    private String getSimulationFileName(Description description) {
-        return description.getTestClass().getSimpleName().toLowerCase() + "/" + description.getMethodName().toLowerCase() + ".json";
+    private SimulationSource getSimulationPath(Description description) {
+        try {
+            final SimulationSource simulationSource = defaultPath(getPerTestMethodSimulation(description));
+            simulationSource.getSimulation(); // to check if it can be loaded - PR to hoverfly to expose check instead would make it nicer
+            return simulationSource;
+        } catch (Exception e) {
+            // Try to load shared per-class simulation
+            return defaultPath(getPerTestClassSimulation(description));
```
tinakurian: should we maybe use the value e so we don't swallow it?

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
bartoszmajsak: Not sure really, this should create key factory, as key alg is built-in RSA. Maybe

```suggestion
throw new IllegalStateException("Something went terribly wrong here, cannot create key factory for " + PUBLIC_KEY_ALGORITHM, e);
```

WDYT?

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
bartoszmajsak:
```suggestion
                new IllegalArgumentException("Header 'X-App' has an invalid value: " + app, iae);
```

would make sense?

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
bartoszmajsak: Funny that IDEA didn't show me that as auto-import... :confused: 

`commons-codec` though is already part of the project - but seems not to be used.

86215721 fixes that

web/src/main/java/io/fabric8/launcher/web/producers/ApplicationTypeBasedProducer.java
```diff
@@ -0,0 +1,29 @@
...
+ * If not present defaults to "launcher"
+ *
+ * @see Application.ApplicationType
+ *
+ * @param <T> type of the bean to look up
+ */
+abstract class ApplicationTypeBasedProducer<T> {
+
+
+    T extractAppBasedImplementation(HttpServletRequest request, Instance<T> instances, Class<T> cls) {
```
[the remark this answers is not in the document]
bartoszmajsak: Right, just keep in mind that `javax.enterprise.util.AnnotationLiteral` comes from the same `javaee-api:7.0` which we already use here.

base-test/src/main/java/io/fabric8/launcher/base/test/hoverfly/LauncherPerTestHoverflyRule.java
tinakurian: sounds good!

base/src/main/java/io/fabric8/launcher/base/identity/RSAPublicKeyConverter.java
tinakurian: looks good!

web/src/main/java/io/fabric8/launcher/web/producers/ApplicationTypeBasedProducer.java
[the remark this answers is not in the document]
gastaldi: Yeah, we are aware that it depends on the CDI API, but depending on Servlet API is another thing :)

core/core-api/src/main/java/io/fabric8/launcher/core/spi/Application.java
bartoszmajsak: Will do when we decide on https://github.com/fabric8-launcher/launcher-backend/pull/557#discussion_r225985580 with @gastaldi

PR #582, opened 2018-10-26
refactor(core): remove dependency on HttpServletRequest in Application

gastaldi:


bartoszmajsak:
I tried to review but it's been merged before I even had a chance to hit "request changes" button. My point being - with opening up conversion method to accept `String` insteand of `HttpServletRequest` error handling is a bit odd - in case of not found instance of `ApplicationType` enum it says 

```java
throw new IllegalArgumentException("Unrecognized application. Header 'X-App' has an invalid value: " + app, iae);
```

which might not longer be the case, as the value to convert is coming from anywhere.

PR #626, opened 2019-01-23
Customize golang booster during create flow

tinakurian:
### Why

Implementing support for "Template-ized Golang Boosters."

### Description

See story: https://openshift.io/openshiftio/Openshift_io/plan/detail/1121

### Example: 
An example modification for the golang-starter rest-http booster is the main.go file which would be modified as follows:

**Before:**
```
package main

import (
	"flag"
	"net/http"
	"os"
	"os/user"
	"runtime"

	"github.com/fabric8-services/fabric8-common/configuration"
	"github.com/fabric8-services/fabric8-common/log"
	"github.com/fabric8-services/fabric8-common/metric"
	"github.com/fabric8-services/fabric8-common/sentry"
	"github.com/goadesign/goa"
	goalogrus "github.com/goadesign/goa/logging/logrus"
	"github.com/goadesign/goa/middleware"
	"github.com/goadesign/goa/middleware/gzip"
	"github.com/golang-starters/golang-rest-http/app"
	"github.com/golang-starters/golang-rest-http/controller"
	"github.com/google/gops/agent"
	"github.com/prometheus/client_golang/prometheus"
	"github.com/prometheus/client_golang/prometheus/promhttp"
)
```
**After:**

```
import (
	"flag"
	"net/http"
	"os"
	"os/user"
	"runtime"

	"github.com/fabric8-services/fabric8-common/configuration"
	"github.com/fabric8-services/fabric8-common/log"
	"github.com/fabric8-services/fabric8-common/metric"
	"github.com/fabric8-services/fabric8-common/sentry"
	"github.com/goadesign/goa"
	goalogrus "github.com/goadesign/goa/logging/logrus"
	"github.com/goadesign/goa/middleware"
	"github.com/goadesign/goa/middleware/gzip"
	"github.com/tinakurian/golang-rest-http/app"
	"github.com/tinakurian/golang-rest-http/controller"
	"github.com/google/gops/agent"
	"github.com/prometheus/client_golang/prometheus"
	"github.com/prometheus/client_golang/prometheus/promhttp"
)
```
**An example can be seen here:** 
_Original booster:_ https://github.com/golang-starters/golang-rest-http
_Openshift create workflow selecting mission and runtime rest-http / Golang:_ https://github.com/tinakurian/TestOpenshiftGolangCreateWorkflow

**Notes:** Unsure about the new file names and package - suggestions are very much welcome!!

### Checklist

- [ ] I followed the contribution [guidelines](https://raw.githubusercontent.com/fabric8-launcher/launcher-backend/master/CONTRIBUTING.md).
- [ ] I created an [issue](https://github.com/fabric8-launcher/launcher-backend/issues/new) and used it in the commit message.
- [ ] I have built the project locally prior to submission with `mvn clean install`.
- [ ] I kept a clean commit log (no unnecessary commits, no merge commits).
- [ ] I am happy with my contribution.


addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Recursively gets files to template.
+     *
+     * @param files
+     *            Files to template
+     */
+    private void getBoosterFiles(File[] directoryContents) {
+        for (File path : directoryContents) {
```
gastaldi: You can use `java.nio.file.Files.walkFileTree` instead. See https://github.com/fabric8-launcher/launcher-backend/blob/master/core/core-impl/src/main/java/io/fabric8/launcher/core/impl/preparers/ChangeArquillianConfigurationPreparer.java#L51 for example

gastaldi:
@tinakurian I have introduced `ProjectileEnrichers`: https://github.com/fabric8-launcher/launcher-backend/pull/627

They basically serve as an extension point to make any necessary changes to the projectile before it is launched (pushed to the github repository, configured in OpenShift, etc)

You should use it instead of `ProjectilePreparer` because you need access to GitHub to gather the user name and in scenarios like downloading the project as a ZIP won't require GitHub access, so your extension point should not be called. The PR in this comment should guarantee that is called appropriately.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBoosterUtility.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Utility class used to assist with customizing golang booster files
+ *
+ */
+public class GolangBoosterUtility {
```
gastaldi: Not sure why this class is necessary, there are plenty of utility functions in commons-io 
 that does the same thing (and it's already available in launcher), like `org.apache.commons.io.FilenameUtils#getExtension`, `org.apache.commons.io.FileUtils#readFileToString(java.io.File, java.nio.charset.Charset)`, among others

tinakurian: This class was meant to just hold helper methods.  I can definitely switch to using getExtension.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,179 @@
...
+     * @return The git organization to replace while templatizing the booster.
+     */
+    @SuppressWarnings("unchecked")
+    private String getGitOrganization(Map<String, Object> boosterData) {
+        HashMap<String, HashMap<String, String>> boosterDataSource = (HashMap<String, HashMap<String, String>>) boosterData.get(SOURCE);
+        HashMap<String, String> boosterGitData = boosterDataSource.get(GIT);
+        String gitOrg = null;
+        try {
+            URL boosterURL = new URL(boosterGitData.get(URL));
+            gitOrg = boosterURL.getPath().split(SEPARATOR)[NumberUtils.INTEGER_ONE];
```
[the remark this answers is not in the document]
tinakurian: I thought i had noticed lang3 as a dependency in the code elsewhere, which is why i thought to use it.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
tinakurian: Good catch, this is better on performance! Will make this change.

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+import io.fabric8.launcher.core.api.Projectile;
+import io.fabric8.launcher.core.spi.ProjectileEnricher;
+import io.fabric8.launcher.osio.projectiles.OsioLaunchProjectile;
+import io.fabric8.launcher.service.git.api.GitService;
+
+/**
+ * Customizes golan booster files.
+ *
+ */
+@ApplicationScoped
```
gastaldi: Make it @Dependent, since GitService is @RequestScoped

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+    private static final String URL = "url";
+    private static final String SOURCE = "source";
+    private static final String START_IMPORT = "import (";
+    private static final String END_IMPORT = ")";
+    private static final String NEW_LINE = "\n";
+    private static final String NO_FILE_EXTENSION = "";
+    private static final String SEPARATOR = "/";
+
+    private Path projectLocation;
+    private Map<String, Object> boosterData;
```
gastaldi: Why the need to store these attributes since they are already available in the `accept` method parameter?

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+    private String gitUser;
+    private Map<File, String> filesToPush = new HashMap<>();
+    private String projectName;
+    private String osioProjectName;
+    private String gitOrg;
+
+    @Override
+    public void accept(Projectile arg0) {
+        OsioLaunchProjectile proj = (OsioLaunchProjectile) arg0;
```
gastaldi: Don't forget to check if the projectile is an instance of `OsioLaunchProjectile` before performing the cast

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * @param boosterData
+     *            The booster data to extract the osio project name from.
+     * @return The osio project name to replace while templatizing the booster.
+     */
+    private String getProjectName(Map<String, Object> boosterData) {
+        if (this.osioProjectName != null) {
+            return this.osioProjectName;
+        }
+        this.osioProjectName = parseBoosterData(boosterData, NumberUtils.INTEGER_TWO);
```
gastaldi: Use the primitive directly, autoboxing is not necessary here

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+
+/**
+ * Customizes golan booster files.
+ *
+ */
+@ApplicationScoped
+public class GolangBooster implements ProjectileEnricher {
+
+    @Inject
+    GitService gitService;
```
gastaldi: Initialize it as part of the constructor of this class to make it easier to test

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * If the file is either a .go file or either makefile or environment file
+     * customize the file by replacing the git organization. Once customized
+     * push the file to the repository.
+     *
+     * @param file
+     *            File to template
+     */
+    private void customizeBoosterFiles(File file) {
```
gastaldi: This method could return the `filesToPush` map (which would be better called `filesToFlush` or `filesToOverwrite`) instead of having it as a class attribute

gastaldi: or rather, why not perform the writing directly? That would be much faster no?

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+        if (isGoFile) {
+            int i;
+            for (i = NumberUtils.INTEGER_ZERO; i < content.size(); i++) {
+                String line = content.get(i);
+                if (line.contains(START_IMPORT)) {
+                    break;
+                }
+            }
+
+            z = i + NumberUtils.INTEGER_ONE;
```
gastaldi: Avoid using this constant. It does autoboxing and hurts performance

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+     * @param file
+     *            File to template
+     */
+    private void customizeBoosterFiles(File file) {
+        boolean pushFile = false;
+        String extension = FileUtils.getExtension(file.getName());
+        boolean isGoExtension = extension.equals(GO_FILE_EXTENSION);
+        List<String> content = new ArrayList<>();
+
+        if (isGoExtension || ((file.getName().equals(MAKEFILE) || file.getName().equals(ASSEMBLE) || file.getName().equals(ENVIRONMENT)) && extension.equals(NO_FILE_EXTENSION))) {
```
gastaldi: This can be a `filter()` in the `Files.walk` method used above

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
...
+     * Extracts the boosters project name in osio.
+     *
+     * @param boosterData
+     *            The booster data to extract the osio project name from.
+     * @param segmentNum
+     *            The segment to return.
+     * @return The booster url segment to return.
+     */
+    @SuppressWarnings("unchecked")
+    private String parseBoosterData(Map<String, Object> boosterData, int segmentNum) {
```
gastaldi: You don't need this method, you can call `getGitRepo()` from the RhoarBooster object inside the Projectile

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
```diff
@@ -0,0 +1,264 @@
+package io.fabric8.launcher.osio.steps.booster;
```
gastaldi: Move this class to a `io.fabric8.launcher.osio.enrichers` package

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
tinakurian: Hi George, thanks for your review. 

Sorry, not done cleaning up, i think it would be better if i maybe pinged you when I'm ready to have it reviewed again?

This list existed prior because i was returning a list to be returned to gitSteps. However, since you delivered the enricher and i no longer utilize anything from gitSteps this doesn't make sense.

I definitely do not want to waste your time and will definitely let you know as soon as I am ready!! 

Thanks again!

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
tinakurian: Ohh okay! That makes sense. Will change this!

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/steps/booster/GolangBooster.java
tinakurian: I should also mention, the ProjectileEnricher you delivered helped clean up a lot of stuff :-) thanks for delivering that so promptly. This became a lot cleaner afterwards - really appreciate it!

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/enrichers/GitUserGoEnricher.java
```diff
@@ -0,0 +1,97 @@
...
+        final String boosterGitOrganization = split[split.length - 2];
+        final String boosterGitRepository = split[split.length - 1];
+        try {
+            Files.walkFileTree(proj.getProjectLocation(), new SimpleFileVisitor<Path>() {
+                @Override
+                public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                    String fileName = file.getFileName().toString();
+                    if (VALID_FILENAMES.contains(fileName) ||
+                            VALID_EXTENSIONS.contains(FilenameUtils.getExtension(fileName))) {
+                        customizeFile(file, gitOrganization, gitRepositoryName, boosterGitOrganization, boosterGitRepository);
```
tinakurian: @gastaldi maybe I am not quite understanding this correctly but the non-existing extension is really only valid if the file is one of "Makefile", "assemble", "environment"

This check looks like it would allow other files such as Jenkinsfile and Dockerfile to pass through?

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/enrichers/GitUserGoEnricher.java
```diff
@@ -0,0 +1,98 @@
...
+    private void customizeFile(Path path, String newGitOrg, String newGitRepo, String oldGitOrg, String oldGitRepo) throws IOException {
+        String contents = new String(Files.readAllBytes(path));
+        int importStartIdx = contents.indexOf(IMPORT_START);
+        if (importStartIdx < 0) {
+            String newContents = contents.replaceAll(oldGitOrg, newGitOrg).replaceAll(oldGitRepo, newGitRepo);
+            Files.write(path, newContents.getBytes());
+        } else {
+            int importEndIdx = contents.indexOf(IMPORT_END, importStartIdx);
+            String importContents = contents.substring(importStartIdx + IMPORT_START.length(), importEndIdx);
+            String newImportContents = importContents.replaceAll(oldGitOrg, newGitOrg).replaceAll(oldGitRepo, newGitRepo);
```
gastaldi: I just realized that the replace should happen using the `org/repo` pattern, otherwise it will replace unwanted content

addons/osio-addon/pom.xml
```diff
@@ -20,6 +20,10 @@
       <groupId>io.fabric8.launcher</groupId>
       <artifactId>launcher-core-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
```
tinakurian: @gastaldi  We actually already have a library available to us which has this capability: String org.codehaus.plexus.util.FileUtils.extension(String filename). Perhaps it's better to use what we already have instead of adding a new library?

gastaldi: That class is part of the Maven integration, I'd rather use commons-io for this instead

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/enrichers/GitUserGoEnricher.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Customizes the import section of .go files as well as the makefile and
+     * environment file to find and replace the git organization value.
+     */
+    private void customizeFile(Path path, String newGitOrg, String newGitRepo, String oldGitOrg, String oldGitRepo) throws IOException {
+        String contents = new String(Files.readAllBytes(path));
+        int importStartIdx = contents.indexOf(IMPORT_START);
+        if (importStartIdx < 0) {
+            String newContents = contents.replaceAll(oldGitOrg, newGitOrg).replaceAll(oldGitRepo, newGitRepo);
```
gastaldi: Don't forget to change this replace to be the same as in Line 92

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/enrichers/GitUserGoEnricher.java
tinakurian: done

addons/osio-addon/pom.xml
tinakurian: sure

addons/osio-addon/src/main/java/io/fabric8/launcher/osio/enrichers/GitUserGoEnricher.java
tinakurian: We cannot do this type of change here. See golang-starters Makefile for example.

gastaldi:
@tinakurian let me know when you want this PR to be merged

sbose78:
Thank you George, let's merge this!

PR #638, opened 2019-01-29
chore: move to quay.io

gastaldi:


gastaldi:
@jmelis I created a PR with the necessary changes to deploy to quay.io, can you verify? (I believe the repositories in quay.io need to be created before this PR is merged)

PR #655, opened 2019-02-08
Trigger build before adding webhook

jarifibrahim:
When a new project is created, two builds are triggered (not always because of a race condition). See openshiftio/openshift.io#3742
1. Because of webhook. We setup webhook before initial commit and that triggers one build (This build doesn't always work because **there is race condition between the config map and the webhook event** @hrishin found this out.)
2. Because of manual trigger build. (this build always work)

In 3001505 we removed the manual build trigger and as a result we had openshiftio/openshift.io#4752 . openshiftio/openshift.io#4752 shows that we have a race condition between the first webhook event and the config map that stores the Jenkins config.

This PR changes the sequence of events so that **we ALWAYS get ONE build**.
The new sequence looks like
1. Create Repository
2. Push changes to repository
3. Push Jenkins Config Map (this should be pushed before the webhook is setup. The config map stores information about Jenkins job that should be triggered on webhook event)
4. Trigger build manually. This will trigger the initial build
5. Setup the webhook on github. This step should always be at the end. This ensures only one build is triggered.

This PR should fix openshiftio/openshift.io#4752 and openshiftio/openshift.io#3742

### DIFF
--- a/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/OsioImportMissionControl.java
+++ b/addons/osio-addon/src/main/java/io/fabric8/launcher/osio/OsioImportMissionControl.java
@@ -85,13 +85,11 @@ public class OsioImportMissionControl implements MissionControl<OsioImportProjec
 
         final BuildConfig buildConfig = openShiftSteps.createBuildConfig(projectile, repository);
 
-        // Create webhook before push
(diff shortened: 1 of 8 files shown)


jarifibrahim:
cc @hrishin @piyush-garg

PR #707, opened 2019-03-18
fix: made cluster connected more robust by also checking if the token in keycloak is valid.

edewit:
fixing: https://github.com/fabric8-launcher/launcher-frontend/issues/1004, https://github.com/fabric8-launcher/launcher-frontend/issues/999

Instead of letting the front-end do n calls ( where n is the number of connected clusters ) to verify if the token stored in keycloak is still valid. We can make the clusters call more "robust" by checking the tokens inline.

- [X] I followed the contribution [guidelines](https://raw.githubusercontent.com/fabric8-launcher/launcher-backend/master/CONTRIBUTING.md).
- [X] I created an [issue](https://github.com/fabric8-launcher/launcher-backend/issues/new) and used it in the commit message. `well to be honest it's all frontend issues`
- [X] I have built the project locally prior to submission with `mvn clean install`.
- [X] I kept a clean commit log (no unnecessary commits, no merge commits).
- [ ] I am happy with my contribution. `why do we ask this question`

### DIFF
--- a/web/src/main/java/io/fabric8/launcher/web/endpoints/OpenShiftEndpoint.java
+++ b/web/src/main/java/io/fabric8/launcher/web/endpoints/OpenShiftEndpoint.java
@@ -23,9 +23,11 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
 
 import com.spotify.futures.CompletableFutures;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.launcher.base.identity.TokenIdentity;
 import io.fabric8.launcher.core.api.security.Secured;
 import io.fabric8.launcher.core.spi.IdentityProvider;
+import io.fabric8.launcher.service.openshift.api.ImmutableParameters;
 import io.fabric8.launcher.service.openshift.api.OpenShiftCluster;
 import io.fabric8.launcher.service.openshift.api.OpenShiftClusterRegistry;
 import io.fabric8.launcher.service.openshift.api.OpenShiftService;
@@ -70,7 +72,20 @@ public class OpenShiftEndpoint {
             Set<OpenShiftCluster> subscribedClusters = clusterRegistry.getSubscribedClusters(securityContext.getUserPrincipal());
             for (OpenShiftCluster cluster : subscribedClusters) {
                 futures.add(identityProvider.getIdentityAsync(authorization, cluster.getId())
-                                    .thenApply(identity -> new ClusterVerified(cluster, identity.isPresent())));
+                                    .thenApply(identity -> {
+                                        if (!identity.isPresent()) {
+                                            return new ClusterVerified(cluster, false);
+                                        }
+                                        final ImmutableParameters.Builder builder = ImmutableParameters.builder().cluster(cluster).identity(identity.get());
+                                        final OpenShiftService service = openShiftServiceFactory.create(builder.build());
+                                        try {
+                                            service.getLoggedUser();
+                                            return new ClusterVerified(cluster, true);
+                                        } catch (KubernetesClientException e) {
+                                            //means that we have an invalid token e.g. cluster got deprovisioned
+                                            return new ClusterVerified(cluster, false);
+                                        }
+                                    }));
             }
             clusters = CompletableFutures.allAsList(futures).get();
         } else {


alien-ike:
### Ike Plugins (test-keeper)

Thank you @edewit for this contribution!



For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin.

web/src/main/java/io/fabric8/launcher/web/endpoints/OpenShiftEndpoint.java
```diff
@@ -70,7 +72,20 @@
...
+                                        if (!identity.isPresent()) {
+                                            return new ClusterVerified(cluster, false);
+                                        }
+                                        final ImmutableParameters.Builder builder = ImmutableParameters.builder().cluster(cluster).identity(identity.get());
+                                        final OpenShiftService service = openShiftServiceFactory.create(builder.build());
+                                        OpenShiftUser user = null;
+                                        try {
+                                            user = service.getLoggedUser();
+                                        } catch (KubernetesClientException e) {
+                                            //means that we have an invalid token e.g. cluster got deprovisioned
```
ia3andy: I guess it's easier to read written like this:
```java
try {
    service.getLoggedUser();
    return new ClusterVerified(cluster, true);
} catch (KubernetesClientException e) {
    //means that we have an invalid token e.g. cluster got deprovisioned
    return new ClusterVerified(cluster, false);
}
```

edewit: I think it depends... updated it nevertheless

ia3andy: @edewit maybe it was subjective, I meant that it was easier at least to me :-)

web/src/main/java/io/fabric8/launcher/web/endpoints/OpenShiftEndpoint.java
```diff
@@ -70,7 +72,20 @@
             Set<OpenShiftCluster> subscribedClusters = clusterRegistry.getSubscribedClusters(securityContext.getUserPrincipal());
             for (OpenShiftCluster cluster : subscribedClusters) {
                 futures.add(identityProvider.getIdentityAsync(authorization, cluster.getId())
-                                    .thenApply(identity -> new ClusterVerified(cluster, identity.isPresent())));
+                                    .thenApply(identity -> {
+                                        if (!identity.isPresent()) {
+                                            return new ClusterVerified(cluster, false);
+                                        }
+                                        final ImmutableParameters.Builder builder = ImmutableParameters.builder().cluster(cluster).identity(identity.get());
+                                        final OpenShiftService service = openShiftServiceFactory.create(builder.build());
```
gastaldi: We should fail this line if the authorization token is invalid. Perhaps the constructor should call getLoggedUser() to test that. 

Better move it to the try/catch block.

edewit: I'm not sure what your asking, do you want the construction of the service to already call openshift so that the "user" of the service knows it's a valid token? What is wrong with the way it is now?

gastaldi: My point is that the `OpenShiftService` is created with an invalid identity, the create must fail

ia3andy: @gastaldi I am not sure you want a request to be made when building a service (neither an exception to be thrown). I think that the call itself should fail, as it actually does..

gastaldi: @ia3andy what's the point of having a Service with an invalid identity set where you can't perform any operations? Or maybe I am missing something

gastaldi: This is what I mean: https://paste.fedoraproject.org/paste/3kcdI9BVmvCDBTvtqkFzjg

ia3andy: @gastaldi having a factory doing a request (or anything beside creating the object it self) is not very nice (IMO).

gastaldi: @ia3andy IMHO the factory must be able to create a valid object. As I mentioned earlier, If the identity is invalid, there is nothing that can be done with the created `OpenShiftService`, so it's better to fail early.

edewit: well it's how the openshiftService works currently and IMO it's nice that it's not doing http requests when you only want to create a service. So I feel changing this behavior should not be part of this PR, this is just to validate if the connected clusters are really connected with a valid token.

gastaldi: Ok, I'll change it anyway

PR #786, opened 2019-06-18
fix: Move fetching the token to the backend.

edewit:
As the frontend can't be trusted with a secret.

### Why

Frondend shouldn't know the secret, this way the secret is never known to the frontend
but the services are still statelss

### Description

The token is fetch and encrypted, what remains to be done is adding the client_id and client_secret into the config.

### DIFF
--- a/services/git-service-api/src/main/java/io/fabric8/launcher/service/git/OAuthTokenProvider.java
+++ b/services/git-service-api/src/main/java/io/fabric8/launcher/service/git/OAuthTokenProvider.java
@@ -1,0 +1,23 @@
+package io.fabric8.launcher.service.git;
+
+import io.fabric8.launcher.service.git.api.GitServiceConfig;
+
+/**
+ * Fetch oauth tokens for git providers
+ */
+public interface OAuthTokenProvider {
+    /**
+     * Get the token for the specified temporary code.
+     * @param code the code to fetch the auth token with
+     * @param config the gitConfig that holds the clientId secret and url
+     * @return encrypted token that can be send back to the server in order to be stateless
+     */
+    String getToken(String code, GitServiceConfig config);
+
+    /**
+     * decrypt an encrypted token.
+     * @param encryptedToken the encrypted token to decrypt
+     * @return the decrypted token
+     */
+    String decryptToken(String encryptedToken);
+}
--- a/services/git-service-impl/pom.xml
+++ b/services/git-service-impl/pom.xml
@@ -31,6 +31,10 @@
       <groupId>net.jodah</groupId>
       <artifactId>failsafe</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.json</artifactId>
+    </dependency>
     <!-- Test dependencies -->
     <dependency>
       <groupId>io.fabric8.launcher</groupId>
--- a/services/git-service-impl/pom.xml
+++ b/services/git-service-impl/pom.xml
@@ -31,10 +31,6 @@
       <groupId>net.jodah</groupId>
       <artifactId>failsafe</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.glassfish</groupId>
-      <artifactId>javax.json</artifactId>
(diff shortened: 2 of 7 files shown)


services/git-service-impl/pom.xml
```diff
@@ -31,6 +31,10 @@
       <groupId>net.jodah</groupId>
       <artifactId>failsafe</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.json</artifactId>
+    </dependency>
```
gastaldi: Use the Jackson library instead

services/git-service-impl/src/main/java/io/fabric8/launcher/service/git/OAuthTokenProviderImpl.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    private String fetchAccessToken(String code, GitServiceConfig config) {
+        JsonObject object = Json.createObjectBuilder()
+                .add("client_id", config.getServerProperties().get("clientId"))
+                .add("client_secret", config.getServerProperties().get("clientSecret"))
+                .add("code", code).build();
+        StringWriter writer = new StringWriter();
+        Json.createWriter(writer).write(object);
```
gastaldi: Use the Jackson library here (see JsonUtils)

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/OAuthTokenProvider.java
```diff
@@ -0,0 +1,23 @@
+package io.fabric8.launcher.service.git;
+
+import io.fabric8.launcher.service.git.api.GitServiceConfig;
+
+/**
+ * Fetch oauth tokens for git providers
+ */
+public interface OAuthTokenProvider {
```
gastaldi: Maybe we can make it an utility class instead? Or will we have different implementations for this interface?

services/git-service-impl/src/main/java/io/fabric8/launcher/service/git/OAuthTokenProviderImpl.java
```diff
@@ -0,0 +1,92 @@
...
+
+@ApplicationScoped
+public class OAuthTokenProviderImpl implements OAuthTokenProvider {
+
+    private final OkHttpClient client;
+
+    private final KeyPair keyPair;
+    private final Cipher cipher;
+
+    OAuthTokenProviderImpl() throws NoSuchAlgorithmException, NoSuchPaddingException {
```
gastaldi: `@Inject OAuthTokenProviderImpl(HttpClient client){...}`

services/git-service-api/src/main/java/io/fabric8/launcher/service/git/OAuthTokenProvider.java
edewit: How would you inject things if it just a static class? It seems all services have one implementation, it would require us to get rid of the `git-service-impl` module?

gastaldi: Sorry, not utility class with static methods, but a single class with `@ApplicationScoped` and constructor injection.

What do you mean by all services having one implementation? AFAIK `GitService` has multiple implementations

edewit: What I mean is all modules are split into 2 one is called `<module>-api` the other `<module>-impl`. Consuming modules only deal with what is in `<module>-api` and they "know" nothing about the `<module>-impl` right? So if we don't want an interface we need to move this implementing class to the `git-service-api` module (or should it be somewhere else)? An example of what I mean is `BoosterCatalogFactory` it's in `core-api` and only has one implementation `RhoarBoosterCatalogFactory` in `core-impl`. So if this reasoning is completely wrong then maybe we need to add it to the contribution guide, with some clear guides to when to do the split and when to create an "utility class"

PR #889, opened 2019-07-10
Fixing setup to ignore user global configuration for "gpgsign = true"

ehsavoie:
### Why

To have tests passing even if the ~/gitconfig contains
[commit] 
             gpgsign = true

Redefine the value for gpgsign to false in the repository used for test

### DIFF
--- a/services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceTest.java
+++ b/services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceTest.java
@@ -41,6 +41,10 @@ import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.lib.ConfigConstants;
+import org.eclipse.jgit.lib.StoredConfig;
+
 public abstract class AbstractGitServiceTest {
 
     private static final String DEFAULT_DESCRIPTION = "The 'best' test repository description with special chars $^¨`\".";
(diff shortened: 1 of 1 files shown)


services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceTest.java
```diff
@@ -272,6 +275,11 @@ public void pushFile() throws Exception {
 
         //When: pushing a README.md file
         final Path tempDirectory = tmpFolder.getRoot().toPath();
+        Git repo = Git.init().setDirectory(tempDirectory.toFile()).call();
```
gastaldi: Better use try-with-resources here in case something bad happens between this line and the `close()` statement

services/git-service-impl/src/test/java/io/fabric8/launcher/service/git/AbstractGitServiceTest.java
```diff
@@ -272,6 +275,11 @@ public void pushFile() throws Exception {
...
+        repo.getRepository().getConfig().setString(ConfigConstants.CONFIG_USER_SECTION, null, ConfigConstants.CONFIG_KEY_EMAIL, "test@redhat.com");
```
gastaldi: Not being picky here, but extracting the Config to a variable may be a good idea :)

quintesse:
@gastaldi I already fixed this with 497b96ed036870b9809823545a4fc363a9677453 , which I _think_ is a better solution in our case.

@ehsavoie 's solution actually writes the value to the repo's config file, which in our case would mean it ends up in the user's repository potentially messing up any GPG setup they might be using locally.

Given the fact that this should only affect devs running the tests I thought it was better to fix this in the code where we push the commits. A better solution might be to have the code ignore any user config settings, but I wasn't able to find a way to do that.

Anyway, closing this PR. Thanks for your time @ehsavoie !

gastaldi:
Ah completely missed that. That makes sense. Thanks @quintesse and @ehsavoie

PR #895, opened 2019-07-12
chore: renamed template references to launcher

gastaldi:
Renamed `launcher-backend` references to `launcher`


gastaldi:
Also this may impact our DSaaS deployments, so let's make sure it gets configured correctly in staging before we put in production

Issue #910, opened 2019-07-24
REGRESSION: Github Oauth is broken


gastaldi:
When deploying the application with GitHub OAuth enabled, the following error occurs when I click on the Authorize button:

`POST http://launcher-launcher123.192.168.42.70.nip.io/launch/github/access_token net::ERR_CONTENT_DECODING_FAILED 405 (Method Not Allowed)`

This is caused because the `/github` endpoint (which used to do a reverse proxy on the Github website) no longer exists.

The work performed by @edewit in https://github.com/fabric8-launcher/launcher-application/pull/786 should have fixed this but apparently this new endpoint is never called

PR #911, opened 2019-07-25
fix: front-end changed to make use of new token fetch on the backend

edewit:
Fixes: #910


edewit:
@gastaldi to test add the clientId and secret in the [git-providers.yaml](https://github.com/fabric8-launcher/launcher-application/pull/911/files#diff-718f83f35b01afedca738e07077a1260R8-R9)

web/src/main/java/io/fabric8/launcher/web/producers/GitServiceProducer.java
```diff
@@ -61,11 +68,12 @@ GitService getGitService(final HttpServletRequest request,
         GitServiceConfig gitServiceConfig = getGitServiceConfig(request);
         GitServiceFactory gitServiceFactory = gitServiceFactories.getGitServiceFactory(gitServiceConfig.getType());
         String provider = gitServiceFactory.getName().toLowerCase();
-        String gitAuthHeader = request.getHeader(GIT_AUTHORIZATION_HEADER);
+        String header = request.getHeader(GIT_AUTHORIZATION_HEADER);
+        String gitAuthHeader = tokenProvider.decryptToken(removeBearerPrefix(header));
```
gastaldi: This breaks existing clients that send the Git token directly in the request. I'd prefer to add a new header indicating that the Bearer token needs to be decrypted before using it (something like `X-Git-Authorization-Encrypted:true`)

gastaldi:
Fixes #910

git-providers.yaml
```diff
@@ -3,6 +3,10 @@
   apiUrl: https://api.github.com
   repositoryUrl: https://github.com
   type: GITHUB
+  serverProperties:
+    oauthUrl: https://github.com/login/oauth/access_token
+    clientId:
```
gastaldi: I am confused, is thie client ID set as an ENV or here?

gastaldi:
Also don't forget to remove unused ENVs from the template in `openshift/template.yaml` and regenerate the unified template

gastaldi:
Looks like the build failed in the frontend module?

web/src/main/java/io/fabric8/launcher/web/producers/GitServiceProducer.java
edewit: removed this completly

git-providers.yaml
edewit: client now loads these settings and no longer uses the env vars

edewit:
@gastaldi still need to update the operator to set the values in the `git-provider.yaml`

git-providers.yaml
```diff
@@ -18,5 +23,10 @@
   apiUrl: https://try.gitea.io
   repositoryUrl: https://try.gitea.io
   type: GITEA
+  clientProperties:
+    giteaOAuthUrl: http://[YOUR-GITEA-URL]/login/oauth/authorize
+    redirectUri: http://frontend-location/fixed
   serverProperties:
     adminUser: admin
+    oauthUrl: https://github.com/login/oauth/access_token
```
gastaldi: This looks weird, maybe leave empty for now?

gastaldi:
```suggestion
    oauthUrl: http://[YOUR-GITEA-URL]/login/oauth/access_token
```

edewit: if you leave it empty it's no longer serializable

PR #962, opened 2019-08-15
feat(openshift): Allow not setting a cluster file

quintesse:
Right now not setting a cluster file result in an error
but since we now allow the user to specify a custom cluster
we should allow not having any clusters defined. It also
makes testing easier when you don't need to configure
everything.

### DIFF
--- a/services/openshift-service-impl/src/main/java/io/fabric8/launcher/service/openshift/impl/OpenShiftClusterRegistryImpl.java
+++ b/services/openshift-service-impl/src/main/java/io/fabric8/launcher/service/openshift/impl/OpenShiftClusterRegistryImpl.java
@@ -49,21 +49,24 @@ public class OpenShiftClusterRegistryImpl implements OpenShiftClusterRegistry {
         if (Objects.toString(apiUrl, "").isEmpty()) {
             // If API or the console URL are not specified, use config file
             String configFile = OpenShiftEnvironment.LAUNCHER_MISSIONCONTROL_OPENSHIFT_CLUSTERS_FILE.value();
-            Objects.requireNonNull(configFile, "Env var " + OpenShiftEnvironment.LAUNCHER_MISSIONCONTROL_OPENSHIFT_CLUSTERS_FILE + " must be set");
-            Path configFilePath = Paths.get(configFile);
-            if (!configFilePath.toFile().isFile()) {
-                throw new IllegalArgumentException("Config file " + configFile + " is not a regular file");
-            }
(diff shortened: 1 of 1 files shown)


services/openshift-service-impl/src/main/java/io/fabric8/launcher/service/openshift/impl/OpenShiftClusterRegistryImpl.java
```diff
@@ -49,21 +49,24 @@ public OpenShiftClusterRegistryImpl(HttpClient httpClient) {
...
+                    defaultCluster = configClusters.get(0);
+                } catch (IOException e) {
+                    throw new IllegalStateException("Error while reading OpenShift Config file", e);
                 }
-                clusters.addAll(configClusters);
-                defaultCluster = configClusters.get(0);
-            } catch (IOException e) {
-                throw new IllegalStateException("Error while reading OpenShift Config file", e);
+            } else {
+                defaultCluster = null;
```
gastaldi: I really wish we had a default value here instead of setting as null. Maybe the OpenShiftConfig class may be used and auto-detect the server URL based on what's in .kube?

gastaldi: Not only null here, but the list of clusters will always be empty in this case

quintesse: Sure, detecting local Kube configs might be nice, but you still might not even have that. So to me that's a possible future improvement. For now I just want to enable the possibility that you can enter a URL + Token in the frontend.

gastaldi:
```suggestion
                defaultCluster = null;
                this.clusters = Collections.emptyList(); 
```

quintesse: That suggestion isn't necessary, isn't it?

I mean we already have this line at the end of the method:

    this.clusters = Collections.unmodifiableSet(clusters);

Which will set it to an empty list if no clusters where read, right?

gastaldi: Ah shoot, missed that, yes, you're right

PR #979, opened 2019-09-03
Fixes #972 New take on #973

### LINKED ISSUE #972: Support OpenShift OAuth with multiple clusters
Each cluster could optionally provide their own OAuth cluster URL

edewit:


web/src/main/java/io/fabric8/launcher/web/endpoints/OpenShiftEndpoint.java
```diff
@@ -64,7 +64,6 @@
     @GET
     @Path("/clusters")
     @Produces(MediaType.APPLICATION_JSON)
-    @Secured
```
gastaldi: Removing `@Secured` will make the `authorizationInstance.get()` fail (or return null, need to test) when no `Authentication` header is specified.

edewit:
not removing it will always result in a a `401`

web/src/main/java/io/fabric8/launcher/web/endpoints/OpenShiftEndpoint.java
gastaldi: According to the spec it would throw `IllegalProductException` but due to a bug in Quarkus it throws NPE 

See https://github.com/quarkusio/quarkus/issues/3829

edewit: not removing it will always result in a a `401`, so it will not work? what do you suggest?

edewit:
@gastaldi could you investigate this further? IMO the removal of the `@Secured` is needed

PR #997, opened 2019-09-30
chore: Updated to latest BCS version

quintesse:


pom.xml
```diff
@@ -19,7 +19,7 @@
     <failOnMissingWebXml>false</failOnMissingWebXml>
     <maven.buildNumber.shortRevisionLength>7</maven.buildNumber.shortRevisionLength>
 
-    <version.booster.catalog.service>58</version.booster.catalog.service>
+    <version.booster.catalog.service>60-SNAPSHOT</version.booster.catalog.service>
```
gastaldi: Snapshots won't build in the CI server

quintesse: Sure, this PR shouldn't be merged as-s. I will update it the moment we have an actual v60 release. Is that okay with you?

PR #1018, opened 2020-04-09
chore(dependencies): Bump Quarkus to 1.3.2.Final

gastaldi:


core/core-impl/pom.xml
```diff
@@ -33,8 +33,8 @@
       <artifactId>commons-text</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-smallrye-context-propagation</artifactId>
```
quintesse, line 37: So Smallrye has become Quarkus-compatible by default?

gastaldi: Yes

gastaldi: More the opposite (Quarkus become Smallrye compatible) :)

quintesse: Aha, ok 👍 😄

PR #1021, opened 2020-04-22
fix: use `useEffect` to update the analytics

edewit:


frontend/packages/launcher-app/src/app/launcher-app.tsx
```diff
@@ -76,20 +75,33 @@ function HomePage(props: {}) {
...
+  const [analytics, setAnalytics] = useState<Analytics>(useAnalytics());
+
+  useEffect(() => {
+    setAnalytics((prev) => {
+      const newAnalytics = trackerToken ? new GoogleAnalytics(trackerToken) : prev;
+      newAnalytics.init();
+      return newAnalytics;
+    });
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [trackerToken]);
```
ia3andy, line 87:
```suggestion
  }, []);
```

Since trackerToken is not part of the "react" scope, you can remove it from the effect dependencies. The empty array will means that this effect will be run only once on the first render, which is what you want ;-)

PR #1028, opened 2020-05-21
use Thorntail 2.5.1.Final-redhat

Ladicek:


Ladicek:
Forgot about this in the previous PR :-)

Ladicek:
Thanks @quintesse! How does this work, BTW? This PR is merged, so the changes should be live now?

quintesse:
It first goes through CI/CD and then end up on staging first (https://launch.prod-preview.openshift.io/), Then if everything is okay it get promoted manually to production whenever we want.

Ladicek:
Ah, gotcha. Thanks. We'll wait then :-)

PR #1031, opened 2020-07-04
Updating to remove default ports

akondasif:
### Why

Default ports are susceptible to vulnerabilities 

### Description

Most cyber attacks occur due to default port usage


akondasif:
Also removed an instance of HTTP without TLS

PR #1035, opened 2020-07-24
Thorntail 2.7

Ladicek:


quintesse:
Thanks for keeping this updated @Ladicek . Although you don't have to if you need the time. This UIO for the creator is not publically accessible anymore anyway.

Ladicek:
Ah it's gone already? I didn't check, and since it's trivial for me to update this, I just did :-) Thanks for info!

quintesse:
Yeah, it was deactivated several weeks ago. BUt of course I didn't throw the code away, they might change their minds again ;-)

